PR #18, opened 2020-12-04
Changqin/shape

sebbader:
Debugging the RDF Schemas


sebbader:
Change log:

* Removed "parent" property in "Referable"
* Added cardinality restriction for Range/max and Range/min properties.
* Correcting the sh:message of the property shape which has ref* as Kind.
* Correcting typos in the error messages of the shapes and descriptions/labels.
* Removed the attribute 'category' from the aas:Property class.

PR #27, opened 2021-03-16
Updating SHACL Shapes

maboeckmann:
Fixing some incorrect prefixes (ssh to sh). Also preventing slashes in prefixes (aas:something/Else) by providing full URIs. This is important for tool support


sebbader (approved):
Bugfix is correct and valid.

PR #29, opened 2021-04-08
Typo in Word RDF in We provide schemas ...

g1zzm0:
Changed "We provide schemas of the Asset Administration Shell for JSON, RFD and XML." to "We provide schemas of the Asset Administration Shell for JSON, RDF and XML."


sebbader:
Hello Nico, thanks for noticing the typo!

PR #41, opened 2021-07-21
adjust the names of classes and attribbutes to get consistent pattern of serialization

changqin26:


changqin26:
The first adjustment is to remove the "Refs/Ref" suffix of the name of classes like submodel, conceptDescription etc.

changqin26:
The second proposal would be that we name all properties in the singular form. Because some properties can only be in the singular form and to achieve consistencies, all other properties are also in the singular form. 

**The properties, which can only be in the singular form.**
ConceptDescription/**isCaseOf** | Reference | 0..*
Formula/**dependsOn** | Reference | 0..*
AssetAdministrationShell/**derivedFrom** | AssetAdministrationShell/derivedForm | 0..1
Extension/**refersTo** | Reference | 0..1
BasicEvent/**observed** | Referable | 1
SubmodelElementCollection/**ordered** | boolean | 0..1

changqin26:
@mristin @JoergWende Could you please take a look at the proposals here and express your opinions?

changqin26:
For the second point,  conceptDescription/isCaseOf and Formula/dependsOn are elements with unlimited cardinality, however, it is almost not possible to express "isCaseOf" and "dependsOn" in plural's, right?

JoergWende:
ok ... what is the question ... or the proposal ?
image

If I remember right - we had _references_ in the past but removed them

schemas/xml/AAS.xsd
```diff
@@ -58,7 +58,7 @@
 			<group ref="aas:hasDataSpecification"/>
 			<element maxOccurs="1" minOccurs="0" name="security" type="abac:security_t"/>
 			<element maxOccurs="1" minOccurs="0" name="derivedFrom" type="aas:reference_t"/>
-			<element maxOccurs="unbounded" minOccurs="0" name="submodelRefs" type="aas:submodelRefs_t"/>
+			<element maxOccurs="unbounded" minOccurs="0" name="submodel" type="aas:submodel_t"/>
```
BirgitBoss, line 61: submodelRefs_t is correct, it is a shared aggregation and shared aggregation are always serialized as reference. But perhaps maxOccurs needs to be "1"

schemas/xml/AAS.xsd
```diff
@@ -73,7 +73,7 @@
 			<element maxOccurs="1" minOccurs="0" name="defaultThumbNail" type="aas:file_t"/>
 			<element maxOccurs="1" minOccurs="0" name="globalAssetId" type="aas:reference_t"/>
 			<element maxOccurs="1" minOccurs="1" name="assetKind" type="aas:assetKind_t"/>
-			<element maxOccurs="unbounded" minOccurs="0" name="billOfMaterials" type="aas:submodelRefs_t"/>
+			<element maxOccurs="unbounded" minOccurs="0" name="billOfMaterials" type="aas:submodel_t"/>
```
BirgitBoss, line 76: submodelRefs_t is correct
But perhaps maxOccurs needs to be "1"

schemas/xml/AAS.xsd
```diff
@@ -118,9 +118,9 @@
 			<element maxOccurs="unbounded" minOccurs="0" name="isCaseOf" type="aas:reference_t"/>
 		</sequence>
 	</complexType>
-	<complexType name="conceptDescriptionRefs_t">
+	<complexType name="conceptDescription_t">
```
BirgitBoss, line 121: conceptDescription_t would be a completely different type than conceptDescriptionRefs. Probably both are needed or one can be removed completly.

BirgitBoss (changes requested):
I think there is a misunderstanding of the concept of "<Referable>Refs_t and <Referable>Ref_t (like SubmodelRefs_t and SubmodelRef_t) in contrast to <Referable>_ t and <Referable>s_t (Submodel_t and Submodels_t). I think the xml serialization is correct like it is. So PR not approved or please explain which of the serialization rules for XML is not followed by the current XML serialization.

changqin26:
Thanks for your explanation, we can close this PR.

PR #57, opened 2021-12-27
Initial Commit for Examples from Details of the Asset Administration Shell Part1 V3.0 RC01

g1zzm0:
Fixed paragraph 6


g1zzm0:
@BirgitBoss @mristin Please assign a reviewer to this pull request

BirgitBoss (changes requested):
We probably should not use "..." in JSON examples etc. any longer. If it is a json file it should be a valid JSON file, shouldn't it?

I miss the chapter 7.4.4 Adding Data Specification Templates for xml mapping

Is it possible to not only link example file but also display it is README file?

I originally expected to not have a separate readme for examples but one per schema (i.e. mapping rules + examples in one file). I feel it is a little bit irritiating that on the JSON readme there is one figure but the other figures are in the example reademe

g1zzm0:
Is chapter 7.4.4 a mapping rule or example?

g1zzm0:
Add chapter 7.4.4

schemas/json/examples/ReferencesJsonExample.json
```diff
@@ -0,0 +1,34 @@
...
+  "conceptDescriptions":[],
+  "assetAdministrationShells":[
+    {
+      "modelType":"AssetAdministrationShell",
+      "submodels":[
+        {
+          "keys":[
+            {
+              "idType":"IRI",
+              "local":true,
```
BirgitBoss: no local in Key since V3.0RC01

schemas/json/examples/ReferencesJsonExample.json
```diff
@@ -0,0 +1,34 @@
...
+        "id":"http://env.com/submodels/S1",
+        "idType":"IRI"
+      },
+      "submodelElements":[]
+    }
+  ],
+  "conceptDescriptions":[],
+  "assetAdministrationShells":[
+    {
+      "modelType":"AssetAdministrationShell",
```
BirgitBoss, line 18: Which variant is correct: the one with name attribute or without?

     "modelType":"AssetAdministrationShell",

Or

      "modelType":{
        "name":"Submodel"
      },

BirgitBoss (changes requested):
Thank you! I removed a local attribute from a Key. With respect to modelType I am not sure what is the correct version, I assume     "modelType":"AssetAdministrationShell" is sufficient.

If so then name needs to be removed from this modelType:

      "modelType":{
        "name":"Submodel"
      },

schemas/json/examples/ReferencesJsonExample.json
g1zzm0: I just took the example from the DotAAS v3.1 and did not make any customization.

schemas/json/examples/ReferencesJsonExample.json
g1zzm0: Table 15 "Exemplary minimal JSON for References" shows this example in the published version of V3.0RC01

BirgitBoss: I see... so this is a bug in V3.0RC01 that we need to fix in V3.0RC02 and we should also in github for V3.0RC01

schemas/json/examples/ReferencesJsonExample.json
BirgitBoss: @Manu3756 May you please have look. It seems to be a bug in V3.0RC01, both variants are used in the examples but I assume only one of them is correct.

Which variant is correct: the one with name attribute or without?

 "modelType":"AssetAdministrationShell",

Or

  "modelType":{
    "name":"Submodel"
  },

schemas/json/examples/ReferencesJsonExample.json
g1zzm0: After all, this is the initial commit. It should be a direct copy from the book "AAS in Details". Every bugfixes and change should be in a separated commit after initial one. Otherwise, it's unclear where, when and why we diverge from the original.

schemas/json/examples/ReferencesJsonExample.json
g1zzm0: In the example from the book, the exact line "modelType" is: "AssetAdministrationShell".

g1zzm0:
For Clarification of the scope, I changed the title for the Pull Request to `Initial Commit for Examples from Details of the Asset Administration Shell Part1 V3.0 RC01 `

g1zzm0:
@Manu3756 can you approve the description of the examples in the README.md is a hard copy of chapter 7.4 - 7.6 of Details of the Asset Administration Shell Part1 V3.0 RC01 ?

g1zzm0:
@BirgitBoss @Manu3756 Changes realized as discussed in UAG AAS

mristin:
@g1zzm0 @BirgitBoss @Manu3756 I'm merging this pull request since I made the schema validation more approachable in #65. There should be a follow-up pull request that fixes the examples so that they conform to the schema.

PR #74, opened 2022-01-16
Change Reviewers in `CONTRIBUTING.md`

g1zzm0:
Previously, we missed to list a couple of reviewers in
`CONTRIBUTING.md`. In this patch, we fix the omissions:
* @mristin is added as a reviewer for the XML, JSON and RDF schemas
  because we will start to use the schema generator [aas-core-codegen]
  at some point, and
* We list Torben Miny with his correct GitHub user name (`@TorbenD`)
  since the current user name is used in GitLab.
  
[aas-core-codegen]: https://github.com/aas-core-works/aas-core-codegen
  
review-requested: @mristin

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -83,11 +83,11 @@ The responsible reviewers are:
 
 | Topic | Path | Reviewer |
 | -------------------- | --------------- | ------------------------- |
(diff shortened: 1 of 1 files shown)


mristin:
LGTM!

PR #93, opened 2022-02-17
Remove all non-RDFS annotations

mristin:
The non-RDFS annotations are not crucial, but are hard to automatically
generate. Since we want to approach the automatic generation as much as
possible, we dispense of all the non-critical annotations.


mristin:
@changqin26 FYI

PR #94, opened 2022-02-17
Change order of classes in  RDF ontology for diffs

mristin:
We need to diff against the generated RDF ontology, so we reshuffle the
classes to match the order of the generated one.


mristin:
@changqin26 FYI

PR #95, opened 2022-02-17
Fix plural/singular in RDF ontology

mristin:
We pluralize all the properties which are aggregations. Some of the
plurals have been missed, and we fix them in this patch.


mristin:
@changqin26 FYI

PR #98, opened 2022-02-17
Fix plurals in JSON Schema

mristin:
Some of the plurals in the property names were omitted by mistake. This
patch fixes the omissions.

### DIFF
--- a/schemas/json/aas.json
+++ b/schemas/json/aas.json
@@ -484,7 +484,7 @@
         },
         {
           "properties": {
-            "values": {
+            "value": {
               "type": "array",
               "items": {
                 "oneOf": [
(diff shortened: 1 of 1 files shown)


schemas/json/aas.json
```diff
@@ -484,7 +484,7 @@
         },
         {
           "properties": {
-            "value": {
+            "values": {
```
Manu3756, line 487: there was a reason to have value without "s" for all submodelElements

mristin: Ok, I'll revert it.

mristin: On a second thought, are you sure? All the other properties are pluralized, also in RDF. Do you happen to remember who introduced that?

mristin:
@Manu3756 could you please have another look?

mristin:
Approved off-GitHub via email.

PR #103, opened 2022-02-18
Fix minor style issues in RDF

mristin:
We previously omitted some of the plurals, whitespaces and written-out
URLs.


mristin:
Approved off-GitHub via email.

PR #110, opened 2022-02-18
Materialize enumerations in RDF

mristin:
Enumeration literals are specific to the enumeration and can not be
compared in common programming languages. Therefore we materialize them
as explicit sets instead of using inheritance relations.


mristin:
@changqin26 FYI

PR #113, opened 2022-03-01
Make whitespace uniform in SHACL

mristin:
We replace tabs with spaces and equalize cases of `  a ` to ` a `.


mristin:
@changqin26 FYI

PR #118, opened 2022-03-02
Make order of min/max count consistent in SHACL

mristin:
We normalize that min always comes first for easier diffs.


mristin:
@changqin26 FYI

PR #121, opened 2022-03-02
Revert `ID` to `Id` in SHACL

mristin:
We revert `ID` to `Id` since the initialism "ID" is considered an
exception in the book.


mristin:
@changqin26 FYI

PR #125, opened 2022-03-02
Remove `skos:` from SHACL

mristin:
The `skos` annotations are not so relevant, so we remove them for easier
diffing.


mristin:
@changqin26 FYI

PR #126, opened 2022-03-02
Fix references in SHACL

mristin:
Some of the attributes pointed to data structures instead of
`Reference`.


mristin:
@changqin26 FYI

PR #127, opened 2022-03-02
Fix subclassing in SHACL

mristin:
We add some of the missing subclasses in SHACL as well as change order
to match the generated SHACL schema.


mristin:
@changqin26 FYI

PR #138, opened 2022-03-02
Review min/maxCount in SHACL against the book

mristin:
We compared the minCount and maxCount against the book and the generated
SHACL schema.

### DIFF
--- a/schemas/rdf/shacl-schema.ttl
+++ b/schemas/rdf/shacl-schema.ttl
@@ -227,7 +227,7 @@ aas:AssetAdministrationShellEnvironmentShape a sh:NodeShape ;
         a sh:PropertyShape ;
         sh:path <https://admin-shell.io/aas/3/0/RC01/AssetAdministrationShellEnvironment/assetAdministrationShells> ;
         sh:class aas:AssetAdministrationShell ;
-        sh:minCount 1 ;
+        sh:minCount 0 ;
         sh:message "(AssetAdministrationShellEnvironment.assetAdministrationShells): At least one <i>assetAdministrationShells</i> attribute having an <i>AssetAdministrationShell</i> entity is required."^^xsd:string ;
     ] ;
     sh:property [
@@ -240,14 +240,14 @@ aas:AssetAdministrationShellEnvironmentShape a sh:NodeShape ;
         a sh:PropertyShape ;
         sh:path <https://admin-shell.io/aas/3/0/RC01/AssetAdministrationShellEnvironment/submodels> ;
         sh:class aas:Submodel ;
-        sh:minCount 1 ;
+        sh:minCount 0 ;
         sh:message "(AssetAdministrationShellEnvironment.submodels): At least one <i>submodels</i> attribute having a <i>Submodel</i> entity is required."^^xsd:string ;
     ] ;
     sh:property [
         a sh:PropertyShape ;
         sh:path <https://admin-shell.io/aas/3/0/RC01/AssetAdministrationShellEnvironment/conceptDescriptions> ;
         sh:class aas:ConceptDescription ;
-        sh:minCount 1 ;
+        sh:minCount 0 ;
         sh:message "(AssetAdministrationShellEnvironment.conceptDescriptions): At least one <i>conceptDescriptions</i> attribute having a <i>ConceptDescription</i> entity is required."^^xsd:string ;
     ] ;
 .
@@ -323,6 +323,7 @@ aas:BlobShape a sh:NodeShape ;
     sh:property [
         a sh:PropertyShape ;
         sh:path <https://admin-shell.io/aas/3/0/RC01/Blob/value> ;
+        sh:minCount 0 ;
         sh:maxCount 1 ;
         sh:message "(Blob.value):Only one <i>value</i> is allowed"^^xsd:string ;
     ] ;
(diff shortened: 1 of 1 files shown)


changqin26 (commented):
Hi @mristin , I think for AssetAdministraionShellEnvironment the minCount for /assetAdministrationShells, /submodels, /conceptDescriptions should be 1 instead of 0.  Reference what JSON schema has:
`"$schema": "https://json-schema.org/draft/2019-09/schema",
  "title": "AssetAdministrationShellEnvironment",
  "$id": "http://www.admin-shell.io/schema/json/V3.0RC01",
  "type": "object",
  "required": [
    "assetAdministrationShells",
    "submodels",
    "assets",
    "conceptDescriptions"
  ],`

mristin:
@changqin26 yes, you are correct! I'll fix that, thanks for spotting it!

mristin:
@changqin26 actually, I'm not sure that the `minCount` should be 1. The `required` in JSON means that a property needs to be defined, but a list can be empty (`[]`). A `minCount` of 1 for, say, submodels would mean that at least *one* submodel must be specified (analogously for assetAdministrationShells *etc.*).

changqin26:
@mristin , I am not an expert on JSON. But according to my understanding of the JSON schema, they use `required` to express `minCount 1`, e,g. AssetInformation has 4 properties, only `assetKind` has the `minCount 1` restriction, so in the `required` area, only `assetKind` is written there. 
`"AssetInformation": {
      "properties": {
        "assetKind": {
          "$ref": "#/definitions/AssetKind"
        },
        "globalAssetId": {
          "$ref": "#/definitions/Reference"
        },
        "specificAssetIds": {
          "type": "array",
          "items": {
            "$ref": "#/definitions/IdentifierKeyValuePair"
          }
        },
        "billOfMaterial": {
          "type": "array",
          "items": {
            "$ref": "#/definitions/Reference"
          }
        },
        "thumbnail": {
          "$ref": "#/definitions/File"
        }
      },
      "required": [
        "assetKind"
      ]`

mristin:
@changqin26 JSON works indeed a bit differently than RDF. The properties of an object are by default optional in JSON. Moreover, a property needs a value, and a value can be either a string, a boolean or a list. An empty list is a valid value for a property and satisfies the `required` constraint in JSON.

Here is an example schema: 

```
{
  "$schema": "https://json-schema.org/draft/2019-09/schema",
  "title": "AssetAdministrationShellEnvironment",
  "$id": "http://www.admin-shell.io/schema/json/V3.0RC01",
  "type": "object",
  "properties": {
    "someProperty": {
      "type": "array"
    }
  },
  "required": [
    "someProperty"
  ]
}
```

and here is an example JSON that satisfies it:

```
{
  "someProperty": []
}
```

You can play with it on https://www.jsonschemavalidator.net/.

Please see http://json-schema.org/understanding-json-schema/reference/object.html#required-properties for more information.

mristin:
P.S. In contrast, RDF considers everything to be a list of values. Hence, what a non-list property is in JSON, needs `maxCount` of 1 in RDF.

changqin26:
@mristin , I think we have a misunderstanding here. `minCount 1 ` is a restriction on the property itself, not on the value of that property. As you mentioned before, " The required in JSON means that a property needs to be defined". `minCount 1` means that a property must be there if you are referring to the class of that property, we don't consider the value of the property there.

mristin:
@changqin26 I asked @g1zzm0 and his colleague Robert for an opinion as I am indeed not certain at all what the correct solution is. I'll come back to you as soon as I receive a response from them, or they write directly here.

mristin:
@Manu3756 Why are all the properties with lists always optional in JSON schema except the lists in `AssetAdministrationShellEnvironment`? I'd make all the properties in `AssetAdministrationShellEnvironment` optional in JSON as well for consistency with the remainder of the JSON schema. This would also resolve the confusion @changqin26  and I face here.

lehmann737:
> @changqin26 I asked @g1zzm0 and his colleague Robert for an opinion as I am indeed not certain at all what the correct solution is. I'll come back to you as soon as I receive a response from them, or they write directly here.

@mristin @changqin26 

While the W3C definition remains somewhat unclear. The "original" definition of TopQuadrant - the guys who defined shacl in the first place  - is somewhat clear:

"sh:minCount and sh:maxCount: Can specify the minimum and maximum number of allowed values for the property, aka the cardinality. If left unspecified, the cardinality is zero-to-many."

sh:minCount works pretty much like owl:ObjectMinCardinality.

mristin:
@lehmann737 the confusion comes from JSON/RDF comparison. In JSON the properties in question are required and have type array, but there are no constraints on the length of those arrays, i.e. the arrays themselves can be empty.

lehmann737:
@mristin 1 actually, cause what you want to express is something like:

<SomeClass> conceptDescriptions <ListOfConceptDescriptions>
with <ListOfConceptDescriptions> eq []  .... or some real content but that's obvious
in this case use minCount 1

if  you want to allow
<SomeClass>  -- that is <SomeClass> not even having the property point to []
... to go without any conceptDescriptions turn minCount to 0 (or omit it as 0 equals to not existent in that  case)

Anyway, I can't really grasp if and where there is the semantic difference in the later application.

PR #139, opened 2022-03-02
Apply postfixes to min/maxCount in SHACL

mristin:
These are minor postfixes which were originally omitted by mistake.


mristin:
@changqin26 you might also want to review this change?

changqin26:
Looks good!

PR #141, opened 2022-03-03
Refer to `SubmodelElement` definition in JSON

mristin:
We replace an in-lined `oneOf` list with the reference to
`SubmodelElement` definition for better readability.


mristin:
@Manu3756 FYI

PR #142, opened 2022-03-03
Simplify redundant `allOf` and `oneOf` in JSON

mristin:
We simply indent the references to the respective definitions for
readability.


mristin:
@Manu3756 FYI

PR #143, opened 2022-03-03
Define `DataElement` in JSON

mristin:
This makes the schema easier to trace from the book even though the
definition for `DataElement` just points to `SubmodelElement`.


mristin:
@Manu3756 FYI

PR #145, opened 2022-03-03
Remove `Content` from data specifications in JSON

mristin:
We change the name of the definitions to follow precisely the book.


mristin:
@Manu3756 FYI

PR #146, opened 2022-03-03
Make environment a separate definition in JSON

mristin:
This makes JSON schema resemble more the RDF+SHACL schemas and makes it
more readable as the preamble is kept short.


mristin:
@Manu3756 FYI

PR #147, opened 2022-03-03
Fix various types in JSON

mristin:
Various definitions and types of properties were invalid w.r.t. the
book. Namely, some properties were defined as data structures instead of
references, and some definitions were missing (such as `Ceritificate`).


mristin:
@Manu3756 FYI

PR #148, opened 2022-03-03
Add missing constraints in JSON

mristin:
We add a couple of missing constraints in JSON schema according to the
book. Notably, we restrict string length to at least 1 character and add
various patterns for strings.

Additionally, we update some of the examples where minor fixes were
possible, and remove the examples where the number of errors was too
high.


mristin:
@Manu3756 FYI

PR #149, opened 2022-03-03
Reformat JSON examples

mristin:
We reformat JSON examples for easier diffing.


mristin:
@Manu3756 @g1zzm0 FYI

PR #150, opened 2022-03-03
Fix data specifications in JSON by the book

mristin:
We went over the tables in the book and fixed the data specifications
accordingly.


mristin:
@Manu3756 FYI. Notably, I followed the inheritance hierarchy w.r.t. `DataSpecificationContent`.

PR #151, opened 2022-03-03
Fix issues in JSON after SHACL review

mristin:
We propagate fixes from a review of SHACL schema to JSON after
controlling them against the book.


mristin:
@Manu3756 FYI

PR #152, opened 2022-03-03
Propagate findings from reviewing JSON schema

mristin:
These are fixes coming after the review of the JSON schema and comparing
it against the generated one.


mristin:
@changqin26 FYI

PR #154, opened 2022-03-03
Fix `ModelTypes` enumeration in JSON

mristin:
We remove the abstract classes from `ModelTypes` enumeration and
re-order the literals to match the order of the automatically generated
JSON schema.


mristin:
@Manu3756 FYI

PR #156, opened 2022-03-03
Add `BlobCertificate` to `ModelTypes` in JSON

mristin:
We list `BlobCertificate` in the enumeration `ModelTypes` since it is a
referable and was unintentionally omitted previously.


mristin:
@Manu3756 FYI

PR #157, opened 2022-03-03
Make properties of environment optional in JSON

mristin:
We make all the properties of `AssetAdministrationShellEnvironment`
optional in JSON to be consistent with other properties given as arrays.
This change also makes the JSON schema more consistent with the SHACL
schema and thus easier to follow and compare.


mristin:
@Manu3756 FYI

PR #160, opened 2022-03-04
Enforce base64 encoding for blobs in JSON

mristin:
We specify that all BlobTypes need to be encoded in Base64.


mristin:
@Manu3756 FYI

PR #112, opened 2022-02-22
Create Initial Commit for draft rc02 schemas

g1zzm0:
Initial Commit for remote branch of draft rc02 without any valid 
content or updated schema of RC02 
Only for CI and documentation purpose

PR #174, opened 2022-04-10
Add README.md for RDF in RC02

g1zzm0:
For the RDF examples, the README.md was added again. This was
necessary because in #112 the explanation was removed together
with the examples during the clean-up.


g1zzm0:
@mristin can you please pull. At the last cleanup for the examples, I unintentionally removed the README for the RDF examples.

PR #178, opened 2022-04-20
Feature/rdf/update to v3.0 rc02

mristin:


mristin:
The schemas got outdated from the book. We submitted the generated schema in the pull request #177 as an alternative. I'm closing this pull request so that we can archive the files for future reads.

PR #183, opened 2022-05-23
Roleback Commit #183

g1zzm0:
We need to reset the commit #183. 
This was necessary because the documentation 
does not correspond to the realization. 
The tests run again after the processing.


g1zzm0:
review-requested: @mristin

PR #198, opened 2022-06-30
Add `Identifiable` literal to `AasIdentifiables`

mristin:
We missed a literal from the final version of the book.

Used:
* [aas-core-codegen 58b16d9]

[aas-core-codegen 58b16d9]: https://github.com/aas-core-works/aas-core-codegen/commit/58b16d9

### DIFF
--- a/schemas/json/aas.json
+++ b/schemas/json/aas.json
@@ -537,6 +537,7 @@
         "Blob",
         "Capability",
         "ConceptDescription",
+        "Identifiable",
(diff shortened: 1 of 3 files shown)


mristin:
Related to https://github.com/aas-core-works/aas-core-meta/issues/131.

mristin:
@sebbader this is a change in enumeration literals, so I suppose you need to re-generate your code as well.

PR #199, opened 2022-07-10
Update to constant sets in aas-core

mristin:
We introduced a major orchestrated change in:

* [aas-core-meta 5c6ac67],
* [aas-core-codegen 6b55dbc], and
* [aas-core3.0rc02-testgen 961195bd]

which introduced constants (notably, constant sets) into the meta-model.

This is reflected in the generated example data.

The notable change is in RDF+SHACL as we remove subsets of enumeration
literals from the definitions. This is necessary, as enumeration
literals can not be compared against each other and had incompatible
URLs. We still have to decide how to deal with subsets of enumeration
literals in RDF+SHACL.

[aas-core-meta 5c6ac67]: https://github.com/aas-core-works/aas-core-meta/commit/5c6ac67
[aas-core-codegen 6b55dbc]: https://github.com/aas-core-works/aas-core-codegen/commit/6b55dbc
[aas-core3.0rc02-testgen 961195bd]: https://github.com/aas-core-works/aas-core3.0rc02-testgen/commit/961195bd


mristin:
@sebbader FYI -- let's have a discussion over email/live call(s) how to deal with subset of enumeration literals in RDF+SHACL.

Issue #218, opened 2022-08-31
Translate examples for RDF


mristin:
The generated examples should be also given in RDF akin to XML and JSON.

mristin:
Closing since addressed in #219.

Issue #215, opened 2022-08-31
xml V3.0RC02 example for environment missing


BirgitBoss:
I love the examples! I miss the (complete and minimal) example for environment

https://github.com/admin-shell-io/aas-specs/tree/draft-V3RC02-schemas/schemas/xml/examples/generated

mristin:
Thanks! This was a bug in my script (the examples are available at https://github.com/aas-core-works/aas-core3.0rc02-testgen/tree/main/test_data/Xml/SelfContained/Expected/environment, I just need to copy them).

mristin:
Closing since addressed in #219.

Issue #214, opened 2022-08-31
Question w.r.t. Enumeration Values Capital Letters?


BirgitBoss:
https://github.com/admin-shell-io/aas-specs/tree/draft-V3RC02-schemas

xml schema:
(as generated as example https://raw.githubusercontent.com/admin-shell-io/aas-specs/draft-V3RC02-schemas/schemas/xml/examples/generated/submodel/complete.xml):

e.g. https://github.com/admin-shell-io/aas-specs/blob/draft-V3RC02-schemas/schemas/xml/examples/generated/basicEventElement/complete.xml

					<direction>OUTPUT</direction>
					<state>OFF</state>

In the specification the enumeration states: 
     input/output or 
      on/off 
as values. Are all enumeration values converted to Capital Letters? Is this a general rules for xml serialization?

For assetKind for example it is not like this:
https://github.com/admin-shell-io/aas-specs/blob/draft-V3RC02-schemas/schemas/xml/examples/generated/assetInformation/complete.xml
		<assetKind>Instance</assetKind>

mristin:
Closing since addressed in #219.

Issue #210, opened 2022-08-31
xml V3.0RC02 Type of Property/value not correct


BirgitBoss:
https://github.com/admin-shell-io/aas-specs/tree/draft-V3RC02-schemas
xml schema:

Type of Propery/value is value „any xsd atomic type as specified via DataTypeDefXsd“ according to specification V3.0RC02 and not string (https://industrialdigitaltwin.org/wp-content/uploads/2022/06/DetailsOfTheAssetAdministrationShell_Part1_V3.0RC02_Final1.pdf)

Note: In previous versions of the spec the type was anySimpleType

Note: for JSON serialization string was accepted as type

mristin:
Closing since addressed in #219.

PR #219, opened 2022-09-01
Update to aas-core-meta, codegen and testgen

mristin:
We update the schemas and the examples using the latest versions of:
* [aas-core-meta 95055a5],
* [aas-core-codegen 295a211], and
* [aas-core3.0rc02-testgen b7fbad50].

This change directly addresses the issues:
* #218,
* #215,
* #214,
* #210, and
* #75.

[aas-core-meta 95055a5]: https://github.com/aas-core-works/aas-core-meta/commit/95055a5
[aas-core3.0rc02-testgen b7fbad50]: https://github.com/aas-core-works/aas-core3.0rc02-testgen/commit/b7fbad50
[aas-core-codegen 295a211]: https://github.com/aas-core-works/aas-core-codegen/commit/295a211


schemas/xml/AAS.xsd
```diff
@@ -703,16 +703,16 @@
         </xs:simpleType>
       </xs:element>
       <xs:element name="valueType" type="dataTypeDefXsd_t"/>
-      <xs:element name="value" type="xs:string" minOccurs="0" maxOccurs="1"/>
+      <xs:element name="value" type="xs:anySimpleType" minOccurs="0" maxOccurs="1"/>
```
BirgitBoss, line 706: anySimpleType is the old version, in V3.0RC02 this was changed (see https://github.com/admin-shell-io/aas-specs/issues/210)  
Type of Propery/value is value „any xsd atomic type as specified via DataTypeDefXsd“ (https://industrialdigitaltwin.org/wp-content/uploads/2022/06/DetailsOfTheAssetAdministrationShell_Part1_V3.0RC02_Final1.pdf)

grafik

schemas/xml/AAS.xsd
```diff
@@ -703,16 +703,16 @@
...
       <xs:element name="valueId" type="reference_t" minOccurs="0" maxOccurs="1"/>
     </xs:sequence>
   </xs:group>
   <xs:group name="range">
     <xs:sequence>
       <xs:group ref="dataElement"/>
       <xs:element name="valueType" type="dataTypeDefXsd_t"/>
-      <xs:element name="min" type="xs:string" minOccurs="0" maxOccurs="1"/>
-      <xs:element name="max" type="xs:string" minOccurs="0" maxOccurs="1"/>
+      <xs:element name="min" type="xs:anySimpleType" minOccurs="0" maxOccurs="1"/>
```
BirgitBoss, line 714: If there would be a type defined as done in meta model (ValueDataType) then there would not be a change in so many places...

BirgitBoss, line 714: If there would be a type defined as done in meta model (ValueDataType) then there would not be a changes in so many places and reviewing would be much easier...

mristin: This is not a good idea for tools which use XSD to generate data structures. At least this was the reasoning why we in-lined the constraints in JSON and RDF schema instead of defining constrained primitive types.

PR #226, opened 2022-09-03
Update CONTRIBUTING.md

BirgitBoss:
Contributing.md was updated for branch V3.0RC02 
(was already accepted and merged with master, 
therefore no new review needed.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -37,9 +37,13 @@ The commit messages follow the guidelines from https://chris.beams.io/posts/git-
 * Use the body to explain *what* and *why* (instead of *how*)
 
 ## Create Pull Request
-After all changes have been committed to your feature branch, a [pull request] has to be created. See [this Github tutorial] for more guidance. 
+After all changes have been committed to your feature branch, a [pull request] (PR) has to be created.
+Every PR should be linked to an issue for tracking.
+See [this Github tutorial] for more guidance. 
 
 [pull request]: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request
+
+[link PR to issue]: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue
 
 ## Pre-Merge Checks
 **Continuous Integration.**
@@ -92,6 +96,6 @@ After the approval the pull request can be merged into the repository. This is d
 **Congratulation.**
 You successfully contributed to the aas-spec repository.
 
-If you are a member of the development team, pleas delete the feature branch you directly created within the aas-specs repository.
+If you are a member of the development team, please delete the feature branch you directly created within the aas-specs repository.
 
 Otherwise, if you are not part of the team and you forked the repository, feel free to delete your fork.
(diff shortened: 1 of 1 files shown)


BirgitBoss:
Second error has nothing to do with this PR since only Contribution.md was changed
 Failed to compile the schema: xml\AAS.xsd
Type 'http://www.w3.org/2001/XMLSchema:dateTimeStamp' is not declared, or is not a simple type.
XML validation failed.

sebbader:
> Second error has nothing to do with this PR since only Contribution.md was changed
> Failed to compile the schema: xml\AAS.xsd
> Type 'http://www.w3.org/2001/XMLSchema:dateTimeStamp' is not declared, or is not a simple type.
> XML validation failed.

That's a really strange one, especially as other PRs pass the test. I have created #226 to reproduce it, and received the same error. However, I was not able to reproduce or debug the script error locally, as I do not have a Win machine available. Can someone else check, too?

mristin:
If not fixed till Wednesday, I'll have a look.

mristin:
@BirgitBoss I suspect the HEAD of your fork is out of sync with the origin.

sebbader:
I used #226 to debug the validation error. Looks like `xs:dateTimeStamp`, `xs:dayTimeDuration`, and `xs:yearMonthDuration` cause the error but e.g. `xs:anyURI`, `xs:base64Binary`, or `xs:boolean` don't affect it. Strange behaviour.

sebbader-sap:
@mristin please take a look at it tomorrow. It seems like these datatypes have not made it into the schema files (=XSD files), even though being part of the standard (=HTML file). If you do not find a better solution, I propose to delete them from https://github.com/admin-shell-io/aas-specs/blob/7ec3e0807bd4a528674a678d1a0d581c45a747ff/schemas/xml/AAS.xsd#L1101 (please note: I also think we should explicitly not delete them from the enum at https://github.com/admin-shell-io/aas-specs/blob/7ec3e0807bd4a528674a678d1a0d581c45a747ff/schemas/xml/AAS.xsd#L968 ).
Of course we then need to write a short sentence in the XML Readme that and why we did this...

mristin:
@sebbader these member types were added in XSD 1.1. If you use a validator which validates using XSD 1.0, it throws an error.

The C# validator we are using is probably based on XSD 1.0. It is only strange that CI did not fail before, when I merged in these changes, and gave a green light.

mristin:
@sebbader I figured out why CI didn't fail. I mistakenly pushed a commit to draft branch (https://github.com/admin-shell-io/aas-specs/commit/7ec3e0807bd4a528674a678d1a0d581c45a747ff#diff-eb36a693a0afb92c36c3cae8975320c12cd7eccaa3cf24b3441ca4131d3cab64) instead of creating a pull request, and CI was set to run only on pull requests.

mristin:
@sebbader I'll update the validator tomorrow.

mristin:
@sebbader great, C# doesn't support XSD 1.1: https://stackoverflow.com/questions/61293382/xsd-1-1-validation-for-both-java-and-net-c

Let's discuss this in email as it needs a broader discussion.

mristin:
@sebbader the issue with CI has been fixed in #230. Please rebase this PR and then you can merge it.

PR #236, opened 2022-09-07
Update Serialization Rules (string + plural)

BirgitBoss:
Three sections updated:
- reformulation of why plural "s" is added in chapter aobut UML properties
- reformulate chapters on string handling: we do not restrict primitive types to string only. 
   Nevertheless the chapter explaining the risks stays in place, only avoiding fomulations like "is not possible"


schemas/json/README.md
```diff
@@ -58,24 +58,35 @@ We explicitly forbid empty JSON arrays to avoid confusion about attributes which
 Namely, an empty array is semantically equal to an omitted attribute (according to the meta-model).
 Thus, the JSON property representing an aggregation attribute must be omitted if the aggregation is empty.
 
-For better readability, we diverge from the meta-model and spell most properties in plural.
+In UML, it is the convention to name associations and aggregations in singular form. The cardinality is to be taking into account to decide on whether there are none, a single or several elements in the corresponding association or aggregation.
```
mristin, line 61: (Also ff.) Please put only one sentence per line for better diffs.

mristin, line 61: s/taking/taken/ or "cardinality indicates ..."

schemas/json/README.md
```diff
@@ -58,24 +58,35 @@ We explicitly forbid empty JSON arrays to avoid confusion about attributes which
 Namely, an empty array is semantically equal to an omitted attribute (according to the meta-model).
 Thus, the JSON property representing an aggregation attribute must be omitted if the aggregation is empty.
 
-For better readability, we diverge from the meta-model and spell most properties in plural.
+In UML, it is the convention to name associations and aggregations in singular form. The cardinality is to be taking into account to decide on whether there are none, a single or several elements in the corresponding association or aggregation. 
+In JSON it is the array that declares the set of elements, therefore we choose the plural name for it. The singular name is used for its descriminator (see section on decriminators).
```
mristin, line 62: An array declares a sequence of elements? I already rewrote this section.

schemas/json/README.md
```diff
@@ -58,24 +58,35 @@ We explicitly forbid empty JSON arrays to avoid confusion about attributes which
...
+Typically the plural name is derived by just adding an "s" to the name.
 For example, ``submodelElements`` instead of ``submodelElement`` in case of [Submodel] class.
 
+If plural form made no sense for a property, we kept it as-is (*e.g.*, `isCaseOf`).
```
mristin, line 66: Duplicate lines?

schemas/json/README.md
mristin, line 61: It seems that you did not update your fork to the origin's branch?

schemas/json/README.md
```diff
@@ -58,24 +58,35 @@ We explicitly forbid empty JSON arrays to avoid confusion about attributes which
...
 If plural form made no sense for an attribute, we kept it as-is (*e.g.*, `isCaseOf`).
 The full list of exceptions is available [as code in aas-core-meta].
 
 [as code in aas-core-meta]: https://github.com/aas-core-works/aas-core-meta/blob/95055a55a8c8f60d75fb48c26eb932ff99945dd2/tests/test_v3rc2.py#L1122
 
-### Primitive attribute value 🠒 JSON string
+### Primitive attribute values
 
-We strictly use only [JSON strings] to represent the attribute values.
+The UML specification uses xsd types. For the mapping of xsd types to JSON types please refer to [Part 2 of the series of the Asset Adminsistration Shell].
```
mristin, line 76: s/xsd/XSD/, also ff.

schemas/json/README.md
```diff
@@ -58,24 +58,35 @@ We explicitly forbid empty JSON arrays to avoid confusion about attributes which
...
+
+There are the following exceptions: For [Property]/value and [Range]/min and [Range]/min the type xs:string is used because the type is only known at run-time. Its type is declared in [Property]/valueType or [Range]/valueType, resp.
+Note: in valueOnly Format of [Part 2 of the series of the Asset Adminsistration Shell] value has the type as declared in Property/valueType.
```
mristin, line 79: " ... has the JSON type .. "?

schemas/json/README.md
```diff
@@ -58,24 +58,35 @@ We explicitly forbid empty JSON arrays to avoid confusion about attributes which
...
 
 [as code in aas-core-meta]: https://github.com/aas-core-works/aas-core-meta/blob/95055a55a8c8f60d75fb48c26eb932ff99945dd2/tests/test_v3rc2.py#L1122
 
-### Primitive attribute value 🠒 JSON string
+### Primitive attribute values
 
-We strictly use only [JSON strings] to represent the attribute values.
+The UML specification uses xsd types. For the mapping of xsd types to JSON types please refer to [Part 2 of the series of the Asset Adminsistration Shell].
+
+There are the following exceptions: For [Property]/value and [Range]/min and [Range]/min the type xs:string is used because the type is only known at run-time. Its type is declared in [Property]/valueType or [Range]/valueType, resp.
```
mristin, line 78: You probably mean "JSON string" instead of `xs:string`?

schemas/json/README.md
```diff
@@ -58,24 +58,35 @@ We explicitly forbid empty JSON arrays to avoid confusion about attributes which
...
 
-[JSON strings]: https://json-schema.org/understanding-json-schema/reference/string.html
 
-This might come as a surprise, given that classes of the meta-model such as [Property] and [Range] allow for boolean and numeric values.
-The reason why we could not map those values to [JSON number] or [JSON boolean] is that the attribute values of the meta-model are based on [XSD types], and not on JSON (see Section [5.7.12 Primitive and Simple Data Types] of the meta-model).
-As [XSD types] do not map to JSON types, we can only represent them as strings.
-We will look more into this problem in the next section.
+### Be aware of possible type mismatches between xsd and JSON types
+
+Type mappings are not trivial. Some sort of type mapping is always needed when using different programming languages and different serialization formats. One way to avoid some JSON type conversion problems would have been to just use [JSON strings] for all primitive attributes values.
```
mristin, line 86: There is no risk of JSON injection attacks that I ever heard of and I couldn'tfind any relate CVE. Please double-check.

mristin, line 86: You mention above that xs:string is always used -- a contradiction?

schemas/json/README.md
```diff
@@ -95,18 +104,17 @@ First, a [JSON number] can only be a concrete numerical value where special case
 Second, the [lexical space] of XSD values is much larger than the lexical space of JSON numbers and booleans.
 For example, both `1` and `true` are valid lexical representations of a true value.
 In contrast, JSON recognizes only `true`.
+We are not considering cases in which a semantic definition would even redefine the semantics of `1` as `false`.
```
mristin, line 107: This is not possible by XSD. XSD is very explicit on both lexical and canonical form. The canonical form of "1" is "True" IIRC.

schemas/json/README.md
```diff
@@ -95,18 +104,17 @@ First, a [JSON number] can only be a concrete numerical value where special case
...
 
 [lexical space]: https://www.w3.org/TR/xmlschema11-2/#lexical-space
 
-Since the semantic of the value is defined outside the model, and might depend directly on the lexical representation, we must represent these values as strings.
-For example, we might define through the attribute [Property/semanticId] that a [Property/value], given as a boolean, distinguishes between `1` and `true` where `1` has a completely different meaning from `true`.
-This distinction would have been lost if we simply serialized the value to a JSON boolean.
+In many cases the semantics of a [Property] or any other [SubmodelElement] is also restricting its value range in case of numerical values. However, when just looking at the [Property/valueType] this information is not given`and the value range seems larger than it actually is allowed to be.
```
mristin, line 111: Unexpected `?

schemas/json/README.md
```diff
@@ -95,18 +104,17 @@ First, a [JSON number] can only be a concrete numerical value where special case
...
 
 [Property/semanticId]: https://www.plattform-i40.de/IP/Redaktion/DE/Downloads/Publikation/Details_of_the_Asset_Administration_Shell_Part1_V3.pdf?__blob=publicationFile&v=10#page=75
 [Property/value]: https://www.plattform-i40.de/IP/Redaktion/DE/Downloads/Publikation/Details_of_the_Asset_Administration_Shell_Part1_V3.pdf?__blob=publicationFile&v=10#page=75
 
 Third, a round-trip conversions XML 🠒 JSON 🠒 XML or RDF 🠒 JSON 🠒 RDF would be lossy.
-This prevents proper testing, and can cause a lot of confusion both for the developers and end-users using, say, GUI editors.
+This makes proper testing hard, and can cause a lot of confusion both for the developers and end-users using, say, GUI editors.
```
mristin, line 117: You can not test an undefined behavior. Since the behavior depends on the underlying JSON libraries, it is practically untestable as long as you don't limit the set of allowed JSON libraries, which is not practically possible.

schemas/json/README.md
mristin: I spent some more thought on this. Actually, allowing JSON numbers and booleans does not protect you against JSON injection as JSON strings are also allowed.

The JSON qualifier is an allOf of types string, boolean and number. JSON injection happens if you do not sanitize the input, so strings would still pass through.

OTOH, if you sanitize the input (i.e. check that `value` and `valueType` match), no JSON injection is possible regardless of whether you use only JSON strings or JSON strings+booleans+numbers.

IMO this sentence is thus wrong w.r.t. the security.

schemas/json/README.md
```diff
@@ -58,24 +58,35 @@ We explicitly forbid empty JSON arrays to avoid confusion about attributes which
...
-[JSON strings]: https://json-schema.org/understanding-json-schema/reference/string.html
 
-This might come as a surprise, given that classes of the meta-model such as [Property] and [Range] allow for boolean and numeric values.
-The reason why we could not map those values to [JSON number] or [JSON boolean] is that the attribute values of the meta-model are based on [XSD types], and not on JSON (see Section [5.7.12 Primitive and Simple Data Types] of the meta-model).
-As [XSD types] do not map to JSON types, we can only represent them as strings.
-We will look more into this problem in the next section.
+### Be aware of possible type mismatches between xsd and JSON types
+
+Type mappings are not trivial. Some sort of type mapping is always needed when using different programming languages and different serialization formats. One way to avoid some JSON type conversion problems would have been to just use [JSON strings] for all primitive attributes values. 
+However, besides loss of semantics and higher effort for programmers there also is the risk of JSON injection attacks.
```
mristin, line 87: After some more thought, this sentence is wrong IMO. First, no semantics is lost when using only strings. It is the other way around (the semantics is lost when clipping XSD datavtypes to JSON). You have both semantic ID and value type attributes. Thus it is a different representation, with no effect to semantics.

Second, it is more work for the programmer. With strings you only convert string -> native type. With anyOf JSON string+boolean+number, you have a much more complex branching logic to decide how to decode to the native type.

schemas/json/README.md
BirgitBoss: Now it is: In UML, it is the convention to name associations and aggregations in singular form. 
The cardinality is to be taken into account to decide on whether there are none, a single or several elements in the 
corresponding association or aggregation. 
In JSON it is best practice to use plural form for array in class properties. 
The singular name is used for its descriminator (see section on decriminators).
Typically the plural name is derived by just adding an "s" to the name.

schemas/json/README.md
BirgitBoss: so XSD

schemas/json/README.md
BirgitBoss: yes, will be corrected

schemas/json/README.md
BirgitBoss: rewrite

schemas/json/README.md
BirgitBoss: so I just write: "there also is a higher risk of using unsanitized JSON input", ok?

schemas/json/README.md
BirgitBoss: You were giving this example originally ....
I will just remove the sentence. Your sentence:

"For example, we might define through the attribute [Property/semanticId](https://www.plattform-i40.de/IP/Redaktion/DE/Downloads/Publikation/Details_of_the_Asset_Administration_Shell_Part1_V3.pdf?__blob=publicationFile&v=10#page=75) that a [Property/value](https://www.plattform-i40.de/IP/Redaktion/DE/Downloads/Publikation/Details_of_the_Asset_Administration_Shell_Part1_V3.pdf?__blob=publicationFile&v=10#page=75), given as a boolean, distinguishes between 1 and true where 1 has a completely different meaning from true. This distinction would have been lost if we simply serialized the value to a JSON boolean."

schemas/json/README.md
BirgitBoss: In GUIs you would just harmonize it. You do not need to display 1:1 what you got.

remove sentence with testing

schemas/json/README.md
mristin: "a higher risk when using unsanitized JSON input"?

schemas/json/README.md
```diff
@@ -58,33 +58,37 @@ We explicitly forbid empty JSON arrays to avoid confusion about attributes which
...
 
-### Be aware of possible type mismatches between xsd and JSON types
+### Be aware of possible type mismatches between XSD and JSON types
 
-Type mappings are not trivial. Some sort of type mapping is always needed when using different programming languages and different serialization formats. One way to avoid some JSON type conversion problems would have been to just use [JSON strings] for all primitive attributes values. 
-However, besides loss of semantics and higher effort for programmers there also is the risk of JSON injection attacks. 
+Type mappings are not trivial. 
+Some sort of type mapping is always needed when using different programming languages and different serialization formats. 
+One way to avoid some JSON type conversion problems would have been to just use [JSON strings] for all primitive attributes values. 
+However, besides loss of semantics and higher effort for programmers there also is a higher risk of using unsanitized JSON input.
```
mristin, line 91: Please see the previous comment. There is equal risk of JSON injenction whether you use only JSON strings _or_ JSON strings+numbers+booleans.

The effort for the programmer is _higher_ (in terms of execution branches, analyzing edge cases etc.) when you use JSON strings+numbers+booleans.

My suggestion is to remove this sentence altogether, and state the real reason which is legacy systems.

mristin: P.S. perhaps a clarification of my comment. The value  is anyOf boolean, number and string. The schema can not enforce that it is a number if `valueType` is, say, `xs:double`. If you don't sanitise the input, the schema allows you to set `value` of a Property to a string even if the value type mandates a number. That finally opens doors for JSON injection.

schemas/json/README.md
sebbader: @BirgitBoss please take a quick look at this line and the '`'.

sebbader (approved):
The changes are fine for me. @BirgitBoss please correct the little typo in line 113, then this PR can be merged.

schemas/json/README.md
```diff
@@ -96,17 +107,15 @@ Second, the [lexical space] of XSD values is much larger than the lexical space
...
-This distinction would have been lost if we simply serialized the value to a JSON boolean.
+In many cases the semantics of a [Property] or any other [SubmodelElement] is also restricting its value range in case of numerical values. However, when just looking at the [Property/valueType] this information is not given`and the value range seems larger than it actually is allowed to be.
 
 [Property/semanticId]: https://www.plattform-i40.de/IP/Redaktion/DE/Downloads/Publikation/Details_of_the_Asset_Administration_Shell_Part1_V3.pdf?__blob=publicationFile&v=10#page=75
 [Property/value]: https://www.plattform-i40.de/IP/Redaktion/DE/Downloads/Publikation/Details_of_the_Asset_Administration_Shell_Part1_V3.pdf?__blob=publicationFile&v=10#page=75
 
-Third, a round-trip conversions XML 🠒 JSON 🠒 XML or RDF 🠒 JSON 🠒 RDF would be lossy.
-This prevents proper testing, and can cause a lot of confusion both for the developers and end-users using, say, GUI editors.
+Third, round-trip conversions XML 🠒 JSON 🠒 XML or RDF 🠒 JSON 🠒 RDF would be lossy.
```
mristin, line 118: Shouldn't it say "are lossy" here as you allow JSON numbers and booleans?

mristin: Also ff. I wrote in perspective of string-only values.

schemas/json/README.md
```diff
@@ -96,17 +107,15 @@ Second, the [lexical space] of XSD values is much larger than the lexical space
 For example, both `1` and `true` are valid lexical representations of a true value.
 In contrast, JSON recognizes only `true`.
 
+
 [lexical space]: https://www.w3.org/TR/xmlschema11-2/#lexical-space
 
-Since the semantic of the value is defined outside the model, and might depend directly on the lexical representation, we must represent these values as strings.
-For example, we might define through the attribute [Property/semanticId] that a [Property/value], given as a boolean, distinguishes between `1` and `true` where `1` has a completely different meaning from `true`.
-This distinction would have been lost if we simply serialized the value to a JSON boolean.
+In many cases the semantics of a [Property] or any other [SubmodelElement] is also restricting its value range in case of numerical values. However, when just looking at the [Property/valueType] this information is not given`and the value range seems larger than it actually is allowed to be.
```
mristin, line 113: Revisiting this sentence. I'd recommend you to remove it, or move it to a different section. The problem is that JSON numbers and boolean _restrict_ the space _more_ than a concept descript might allow for, not less!

For example, if a concept is a 64-bit hash ID, it would make sense to represent it as `xs:long`. Since JSON numbers are used, it must be represented as `xs:string`. 

Or a concept some sensor value. If infinity or NaN values are possible to be sent from the sensor, the value type again has to be `xs:string` and the value represented as JSON string, not `xs:double`, as JSON numbers do not allow for infinity and NaN.


I'd recommend you to make examples in the readme as this behavior is quite opaque, and not self-evident at all. I suppose picking a small number, large number, a NaN and -0 would suffice, and show how Property would look like in those cases.

BirgitBoss: This is not correct, since the Property/ValueType is derived from the concept description und not vice versa.
It is always a two-step mapping:
- Mapping from Concept Descirption to xsd
- Mapping from  XSD to target language

For JSON we do habe a mapping table that is the master for the mapping. So if changes are needed the change requests need to be addressed to this mapping table in part 2.

mristin: Please re-visit this part of my comment:
```
For example, if a concept is a 64-bit hash ID, it would make sense to represent it as xs:long. Since JSON numbers are used, it must be however represented as xs:string.
```

So CD -> xs:long. Xs:long does not fit in JSON number (2^63-1 > 2^52). Therefore xs:long's are represented as JSON strings? Or only if the concrete value is larger than 2^52?

BirgitBoss: As I stated: this is a request to the mapping table not to the serialization rule themselves that just states that this table is used.

BirgitBoss: additionallty added comment in issue on source of mapping tabel, see below

schemas/json/README.md
BirgitBoss: In the meantime the complete sentence was removed but you are right, should have been reformulated

BirgitBoss: I did not mention this, if it is a copy/paste error

schemas/json/README.md
BirgitBoss: Sentence removed in the mean-time

schemas/json/README.md
BirgitBoss: The JSON injection risk comes into play with ValueOnly Format. With Normal Output Format the value is xs:string. See exceptions from mappings.

schemas/json/README.md
BirgitBoss: chapter was removed

BirgitBoss:
With respect to INF, -INF, NaN see also other options like discussed in https://github.com/OpenManufacturingPlatform/sds-bamm-aspect-meta-model/issues/175#issuecomment-1240864403, especially Option 4 
Option 4: For Properties with a xsd:float/xsd:double data type, the JSON payload may contain instead a discriminator property which indicates that the value is not numeric but actually INF/-INF/Na

BirgitBoss:
Review finished

PR #264, opened 2023-04-19
Update the naming scheme of RDF and SHACL

### LINKED ISSUE #263: Inconsistent naming of DataSpecificationIec61360
Currently, the V3-schemas for [JSON](https://github.com/admin-shell-io/aas-specs/blob/6f7d5b1e386d950cbd02c477d80065fb2e95fa99/schemas/json/aas.json#L363) and [XML](https://github.com/admin-shell-io/aas-specs/blob/6f7d5b1e386d950cbd02c477d80065fb2e95fa99/schemas/xml/AAS.xsd#L212) use the spelling `DataSpecificationIec61360`. However, the schemas in [RDF and SHACL](https://github.com/admin-shell-io/aas-specs/tree/master/schemas/rdf) use the spelling [`DataSpecificationIEC61360`](https://github.com/admin-shell-io/aas-specs/blob/6f7d5b1e386d950cbd02c477d80065fb2e95fa99/schemas/rdf/rdf-ontology.ttl#L438). This difference is not only inconsistent and therefore impractical in itself. It ripples on and breaks the AAS4J serializers. 

Therefore, I suggest to adjust the naming to `DataSpecificationIec61360` in RDF and SHACL as well.

s-heppner:
Previously, the naming scheme of the RDF and SHACL schemas were 
inconsistent with the JSON and XML schemas in their naming of 
abbreviations.

This patch fixes the naming by applying the same rules on the RDF 
and SHACL schemas. Furthermore we update the examples.

This fixes #263.


schemas/rdf/examples/generated/AdministrativeInformation/maximal.ttl
```diff
@@ -5,7 +5,6 @@
 @prefix xs: <http://www.w3.org/2001/XMLSchema#> .
 
 <something_142922d6> rdf:type aas:AssetAdministrationShell ;
-    <https://admin-shell.io/aas/3/0/Referable/idShort> "something_783819f1"^^xs:string ;
```
sebbader-sap, line 8: It looks like that the `idShort` property disappeared in all the examples. Can we get them back?

schemas/rdf/rdf-ontology.ttl
```diff
@@ -1,2058 +1,2058 @@
...
+@prefix aas: <https://admin-shell.io/aas/3/0/> .
+@prefix owl: <http://www.w3.org/2002/07/owl#> .
+@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
+@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
+@prefix xs: <http://www.w3.org/2001/XMLSchema#> .
+@base <https://admin-shell.io/aas/3/0/> .
+
+<https://admin-shell.io/aas/3/0/> rdf:type owl:Ontology ;
+    owl:versionInfo "V3.0-work-in-progress" ;
+    rdfs:comment "This ontology represents the data model for the Asset Administration Shell according to the specification 'Details of the Asset Administration Shell - Part 1 - Version V3.0-work-in-progress'."@en ;
```
sebbader-sap, line 10: Can we change bot the `versionInfo` and the comment to "V3.0" and "Specification of the Asset Administration Shell: Part 1"?

sebbader-sap (changes requested):

schemas/rdf/examples/generated/AdministrativeInformation/maximal.ttl
mristin: No, that is unfortunately not possible as ID-shorts are optional properties. Their presence in the previous examples was due to the bug I introduced when I misinterpreted that identifiables *must* have ID-shorts, while in reality they do not.

For some history & future reference, see the following pull requests:
* https://github.com/aas-core-works/aas-core-codegen/pull/365: I introduced the type strengthening in aas-core-codegen after introducing the constraint in aas-core-meta with https://github.com/aas-core-works/aas-core-meta/pull/256/files and https://github.com/aas-core-works/aas-core-meta/pull/261.
* After realizing that I misinterpreted Constraint AASd-117, I reverted the changes in https://github.com/aas-core-works/aas-core-meta/pull/264 and https://github.com/aas-core-works/aas-core-codegen/pull/379.

The logic for `maximal` tests in testgen sets all the required properties of the relevant object (an instance of `AdministrativeInformation` in your example), but embeds it in a minimal environment for readability. When I erroneously switched `IdShort` to required in https://github.com/aas-core-works/aas-core-meta/pull/261, testgen obviously included `IdShort` in the minimal examples of the wrapping objects around the instance of `AdministrativeInformation`.

Now that I reverted back to the (correct) specification in https://github.com/aas-core-works/aas-core-meta/pull/264, testgen generates the minimal wrapping without `IdShort` as before.

schemas/rdf/rdf-ontology.ttl
mristin: I fixed it in https://github.com/aas-core-works/aas-core-meta/pull/268.

I changed the comment to a more general one, so that it survives future versions of the specs: https://github.com/aas-core-works/aas-core-codegen/pull/381.

In particular, see:
* https://github.com/aas-core-works/aas-core-codegen/pull/381/files#diff-4a6f81ac34dc84c7116ed5bc6285a39d2f4d8a12344526951b77b681f285a672 , and 
* https://github.com/aas-core-works/aas-core-codegen/pull/381/files#diff-8c8150f399182260bfcf8beca1530bf28db4150bbaf6dae9169f0d9c8eeaacb3.

mristin:
@sebbader-sap please double-check my answers & open the conversations again if necessary.

@s-heppner the schemas and the test data have been regenerated in aas-core-codegen and testgen repositories so you can update this pull request.

schemas/rdf/examples/generated/AdministrativeInformation/maximal.ttl
sebbader-sap: > After realizing that I misinterpreted Constraint AASd-117, I reverted the changes in

I see, the formulation of `AASd-117` and its note is in Part 1 V3.0 a little bit different than I remembered it. The new state is therefore indeed correct.

sebbader-sap (approved):

s-heppner:
I updated the schemata from aas-core, using 

- [aas-core-meta 8f18a8c](https://github.com/aas-core-works/aas-core-meta/commit/8f18a8c) and
- [aas-core-codegen 0e16f63](https://github.com/aas-core-works/aas-core-codegen/commit/0e16f63).

To my understanding, this should make this PR ready to be merged.

sebbader-sap (approved):

sebbader-sap:
@s-heppner both this PR and #268 look good to me - but before merging let's quickly agree on the release steps right after. We should immediately go for a `V3.0.7` release, right? If so, we need to prepare an update of the release section in the README including the new release links.

s-heppner:
Imo, we should also definitely include #269 in the release, since that fixes a major bug in the XSD. 

Other than that, I don't have a strong opinion on release numbering (though 3.0.7 seems sensible). Is there anyone else we should clarify our further action with?

PR #309, opened 2023-12-12
Validate submodel reference in JSON Schema

mhrimaz:
Check submodel references to be only `ModelReference<SubmodelReference>`

### DIFF
--- a/schemas/json/aas.json
+++ b/schemas/json/aas.json
@@ -139,7 +139,7 @@
             "submodels": {
               "type": "array",
               "items": {
-                "$ref": "#/definitions/Reference"
+                "$ref": "#/definitions/ModelReferenceToSubmodel"
               },
               "minItems": 1
             },
@@ -800,6 +800,23 @@
         "value"
       ]
     },
+    "SubmodelKey": {
+      "type": "object",
+      "allOf": [
+        {
+          "$ref": "#/definitions/Key"
+        },
+        {
+          "type": "object",
+          "properties": {
+            "type": {
+              "const": "Submodel"
+            }
+          },
+          "required": ["type"]
+        }
+      ]
+    },
     "KeyTypes": {
       "type": "string",
       "enum": [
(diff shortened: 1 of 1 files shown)


mristin:
@mhrimaz Thanks for the PR!

@sebbader-sap @s-heppner maybe we should collect all these patches and re-write them  as Python code so that they can be re-applied to new versions as well?

mristin:
@sebbader-sap the constraint has been in aas-core-meta since forever, but we can not transpile it to JSON schema. At least, I don't know how.

schemas/json/aas.json
```diff
@@ -1207,6 +1224,34 @@
...
+      "type": "object",
+      "allOf": [
+        {
+          "$ref": "#/definitions/Reference"
+        },
+        {
+          "type": "object",
+          "properties": {
+            "type": {
+              "const": "ModelReference"
```
BirgitBoss, line 1237:
```suggestion
              "const": "Submodel"
```

mhrimaz: Thanks for the review @BirgitBoss.
The schema that I provided only enforces that the reference should be `ModelReference` and not `ExternalReference` with only one `Key`. So for enforcing that the key should refer to a `Submodel` then, we should further extend it.

In any case the fix should be in the aas-core schema generator.

s-heppner:
We truly thank you for your input, @mhrimaz!

However, we sadly cannot allow changing the schema files by hand, since they are generated using [aas-core-works](https://github.com/aas-core-works). 
That particular constraint is implemented in [aas-core-meta](https://github.com/aas-core-works/aas-core-meta) (meaning it appears for example in the generated SDKs), however cannot be transpiled into the JSON schema with reasonable effort, so that we decided not to implement that particular generation function. 
If you're willing to, you're more than welcome to contribute to the schema generation functionality at [aas-core-codegen](https://github.com/aas-core-works/aas-core-codegen). 

@mristin We decided against the semantic patching approach you mentioned due to not wanting to introduce another layer of code that needs to be maintained.

mristin:
I'll close the pull request for now, but feel free to re-open it in the future.

PR #317, opened 2023-12-18
Resolve Issues V3.1

### LINKED ISSUE #287: [Pt.1] Remove EntityType
The Entity-SME currently requires to use the EntityType attribute which can bei either "CoManagedEntity" or "SelfManagedEntity", signifying whether the referred Asset has its own AAS.

image

Here's the issue:

Someone referring to an asset (for instance in a BOM) **will not always know** if the referred Asset has an AAS or not. There's no way to be sure that an Asset has no AAS. That means everyone modeling an Entity is forced to make a definitive statement which (in most cases) can not be made definitively.

### LINKED ISSUE #294: Extension of the AssetKind class by a further entry
Extension of the AssetKind class by a further entry **"Role"** or **"TechnicalPlaceholder"**.

Background: In the context of engineering, resources are specified. At a later point in time, these are assigned an article (type) that fulfils the requirements of the equipment. The "TechnicalPlaceholder" is therefore a form of requirement definition that is indispensable in engineering. This workflow is typical for engineering, which is an essential starting point for a new asset administration shells on machine or plant level. 
Especially during the life cycle of a machine, types can be exchanged. As a rule, the equipment definition (**"TechnicalPlaceholder"**) is kept more general, so that different types from different manufacturers can be used for the defined equipment at a later date.

image

Relation between TechnicalPlaceholder, Type and Instance:
image

### LINKED ISSUE #295: Relaxation of idShort restrictions (Constraint AASd-002)
# What

_idShort_ is currently very restricted defined in the specification:

>idShort of Referables shall only feature letters, digits,
underscore ("_"); starting mandatory with a letter, i.e. [a-zA-Z][a-zA-Z0-9_]*.

idShorts like "min-temperature-value" or "modbus:function" are not allowed. 

idShorts have a variable-like character. However, I cannot follow why only "_" is allowed as a special character. This complicates the reusage of existing names (that are also variable-like characters), e.g., from standards that uses prefixes.

I tried to understand, where the restrictions come from and I received two resealable answers so far.:

1. "should not cause conflicts with the idShort path approach of the REST interface."
2. "variables in programming languages also do not support special characters"

# Mitigation

Point 1 is not really a justification for this restriction, URL path allows many special characters such as “-”, “$”, “:” etc
Point 2 is more justifiable. However, it is hard to generalize this, since some programming languages allow “$” in variable names. But seriously questions: Why should idShort names identical be reflected as variable names in a programming language? Which tool/lib is doing this? AAS comes with its own serialization approaches. It makes more sense to reflect the serialization model in a programmatically manner (if needed).

# Proposal
It makes sense, that idShorts have a variable-like character, however, more flexibility of the idShort values would be desirable, e.g., to adhere existing name conventions that uses also “-” and “:” in the name value.  

Proposal 1: Allow more special characters such as “-” and “:”

and/or

Proposal 2: Allow “%” --> allows URL encoding (all special characters can be reflected)

Note: Proposal 1 and 2 are backward compatible!

### LINKED ISSUE #296: Inconsistency in the cardinality of DataSpecification in EmbeddedDataSpecification
As per the "Specification of Asset Administration Shell Part1 - Metamodel", section 7.2.5, figure 58 "Realization of Embedded Data Specification", the DataSpecification is an optional parameter in EmbeddedDataSpecification. However, in the [SwaggerHub definitions ](https://app.swaggerhub.com/domains/Plattform_i40/Part1-MetaModel-Schemas/V3.0.1#/components/schemas/EmbeddedDataSpecification)or in ["aas-core-works" HTML rendered pages](https://aas-core-works.github.io/aas-core-meta/v3/EmbeddedDataSpecification.html), the DataSpecification is a mandatory parameter. The generated "aas-core-sdk (c#)" also treats it as a mandatory parameter. So, the question here is, if the DataSpecification should be treated as optional or mandatory parameter in EmbeddedDataSpecification?  

Figure 58 from "Specification of Asset Administration Shell Part1 - Metamodel"
MicrosoftTeams-image (2)

### LINKED ISSUE #299: File.value / PathType is inconsistent / too strict
PathType is defined on page 97 as:

Identifier
Note: for any string conformant to RFC8089, the "file" URI scheme (for relative and absolute file paths) applies.

Exmaples:
./Specification.pdf
file:c:/local/Specification.pdf
file://host.example.com/path/to/file

This is more strict than it was in Version 2 (it was a simple string) and inconsistent, since the example ./Specification.pdf is not valid according to RFC8089.

As discussed in Workstream AAS, the specification should be changed to:

`URL, based on RFC-3986 / RFC-3987, but more recently defined as living standard by WHATWG, as suggested by W3C: https://url.spec.whatwg.org/`

In the examples we should also list http / ftp  / custom-scheme examples - since all of them are considered valid.

### LINKED ISSUE #306: Change length of data type Identifier from 2000 to 2048 Char

Change length of data type Identifier from 2000 to 2024 Char (Correction: 2048)

### LINKED ISSUE #312: [BUG] use xsd 1.0 data types
**Describe the bug**
In *DataTypeDefXsd* links to xmlschema11-2, i.e. xsd V1.1 are contained. However, the decision is to use xsd 1.0. See also Figure 54

The xsd schema is using 1.0

**Where**
substitute xmlschema11-2 with xmlschema2 links

### LINKED ISSUE #313: [BUG] Editorial inconsistencies
**Describe the bug**
This issue is serving to track editorial inconsistencies that have impact on documentation only.

* Explanation of HasKind/kind – not type but template
* MultiLanguageNameType inconsistent description, maxLength=128 (schemata were correct)
* Referable: Referable elements can be referenced via the idShort. For details on referencing, see Clause 5.3.9. This is not true for elements within SubmodelElementList. E.g. Referables are element that can be unambiguously referenced within the context of an Identifiable or
Referables are elements within the information model that have the attribute “idShort” or are numerically identifiable
* Metamodel of HasKind shall be part of chapter “HasKind”, not at the end of previous chapter
* Non-identifiable referable elements can be referenced. However, this requires the context of the element. A referable that is not identifiable has a short identifier (idShort) that is unique just in its context, its name space.  this does not hold for SubmodelElementList
A Referable not being Identifiable has either a short identifier (idShort) or a numerical identifier that is unique just in its context, its name space.
also here: Referable elements can be referenced via the idShort. Or a numerical identifier
also here: Note1 : an element that is referable by its idShort. This ID is not globally unique. This ID is unique within the name space of the element.
* remove `<<DataType>>` from Annex UML since not used any longer
* Link to Now ECLASS has mapping https://eclass.eu/support/technical-specification/data-model/datatype-to-xsd-mapping again, put to Part 3a? available again
* the display name is selected in the following order, if available:  could and NOTE
* Examples for ModelReferences: add information for the grammar used to describe the examples

BirgitBoss:
Specification updated conformant to the accepted changes to V3.1

Major Changes:

* New value "Role" to enumeration AssetKind 
  closes #294
* PathType: support complete URI scheme for files RFC3986/RFC3987
   closes #299
* data type  Identifier: change length from 2000 to 2024 characters
   closes #306
* Referable/idShort and Constraint AASd-002: now also allows 
   hyphens to be included in name 
   closes #295
* Entity/entityType and Constraint AASd-014: now optional 
   closes #287

Minor Changes:

* editorial changes

Bugfixes:

* (Editorial) MultiLanguageNameType inconsistent description, 
    maxLength=128 (schemata were correct) 
    part of #313
* (Editorial) HasKind/kind explanataroy text: use template, not type
    part of #313
* (Editorial) DataTypeDefXsd links now correctly to xmlschema11-2, 
    not to xsd V1.1 
   closes #312
*  (Editorial) make mbeddedDataSpecification/dataSpecification
     mandatory  (as implemented in schemata) 
     closes #296
* (Editorial) Text was fixed w.r.t. Referables and exception with 
   SubmodelElementList 
* remove `<<DataType>>` Stereotype from UML Annex 
   since not used any longer
  (part of #313)
  
closes #313


documentation/IDTA-01001/modules/Constraints/constraints.adoc
```diff
@@ -8,12 +8,12 @@ SPDX-License-Identifier: CC-BY-4.0
 ////
 
 // Constraints
-:aasd002: pass:q[[underline]#Constraint AASd-002:# _idShort_ of __Referable__s shall only feature letters, digits, underscore ("_"); starting mandatory with a letter, i.e. [a-zA-Z][a-zA-Z0-9_]*.]
+:aasd002: pass:q[[underline]#Constraint AASd-002:# _idShort_ of __Referable__s shall only feature letters, digits, hyphen ("-") and  underscore ("_"); starting mandatory with a letter, and not ending with a hypen or underscore, i.e. ^[a-zA-Z][a-zA-Z0-9_-]*[a-zA-Z0-9_]+$.]
```
s-heppner, line 11: > ^[a-zA-Z][a-zA-Z0-9_-]*[a-zA-Z0-9_]+$.]

Good eye!

documentation/IDTA-01001/modules/ROOT/IDTA-01001.adoc
```diff
@@ -3634,16 +3642,19 @@ _Identifier_
...
 
 ====
-Note: for any string conformant to link:https://datatracker.ietf.org/doc/html/rfc8089[RFC8089], the "file" URI scheme (for relative and absolute file paths) applies.
+Note: URL, based on [RFC-3986](https://www.rfc-editor.org/rfc/rfc3987) /[RFC-3987](https://www.rfc-editor.org/rfc/rfc3987), but more recently defined as living standard by WHATWG, as suggested by W3C: https://url.spec.whatwg.org/
 ====
 
 
 a|
-./Specification.pdf
+\./Specification.pdf
```
s-heppner, line 3650: What's the reason for the backslash `\` here? As far as I know, only Windows machines allow for backslashes in directory paths.

s-heppner (commented):
This makes the reviewing process very smooth.

documentation/IDTA-01001/modules/Constraints/constraints.adoc
morhof123: Small nitpick, currently this allows the use of an underscore as a last character, when in the sentence before it is stated, that it shouldn't end with an underscore.

BirgitBoss: Thank you for the finding. Unfortunately, we need to be backward compatible and it seems we allowed "_" as last char before. So we need to fix the text.

documentation/IDTA-01001/modules/ROOT/IDTA-01001.adoc
BirgitBoss: otherwise the . is not displayed in asciidoc - at least this was this way in my editor

BirgitBoss: \ removed, thank you

PR #329, opened 2024-01-05
Improve documentation structure for antora

zrgt:
- Moved all files from named-module "Shared" to "ROOT", 
as it is easier to handle all adoc filkes in one module rather than two.
- Created "partials" directory for adoc. files, which are not pages,
but rathe only include some contents.
See https://docs.antora.org/antora/latest/partials-directory/
- Removed index.adoc as not needed for antora
- Fixed section titles in 
IDTA-01xxx_TermsDefinitionsAbbreviations.adoc


BirgitBoss:
index.adoc is useful in antora: https://docs.antora.org/antora/latest/page/start-page/

PR #348, opened 2024-02-06
Update index

monsieuremre:
Having the giant book cover image breaks the website. Instead dynamically recreate the title in asciidoc.


BirgitBoss:
I prefer to start with the cover of the document.

PR #349, opened 2024-02-06
Update antora.yml

monsieuremre:
See: https://github.com/admin-shell-io/aas-specs-api/pull/225


documentation/IDTA-01001/antora.yml
```diff
@@ -1,6 +1,6 @@
 name: IDTA-01001
-title: DRAFT - Specification Asset Administration Shell - Part 1 Metamodel
+title: Part 1: Metamodel
```
BirgitBoss, line 2: The Draft is essential: we will remove it short before release

PR #353, opened 2024-02-14
Update antora.yml

monsieuremre:
We need to add " around the title. Because the title includes a semi colon: the yaml file thinks it's another entry while it is part of the title, and the website fails to build because of falsely formatted yml file.


monsieuremre:
Now it looks like this.

https://rwth-iat.github.io/aas-specs-antora/IDTA-01001/snapshot/index.html

If you want we can still add the cover page, but we have to declare the dimensions so that it wont break the look.

They can be made to look like this.

https://rwth-iat.github.io/aas-specs-antora/index/dev/index.html

PR #354, opened 2024-02-15
Adopt termsIEC 63278-1:2023, Matching References

### LINKED ISSUE #347: Matching Strategy for Reference: what about referredSemanticId?
A Reference has a type, a referredSemanticId and a list of keys. In the chapter "Matching Strategies - Matching Algorithm for References" the referredSemanticId is not mentioned. Which impact does the referredSemanticId has on the matching algorithm?

The Note `Note: if the key types are not identical although all key values follow the correct order of the key chain, then at least one of the references is buggy and a warning should be issued.` is true for match model-ref against model-ref and external-ref against external-ref but not for external-ref against model-ref. This should be added

### LINKED ISSUE #350: Matching Strategy: Outdated equivalence of ModelReference[Submodel] and GlobalReference with same key
[Page 36 of the v3.0 Part 1 specification](https://industrialdigitaltwin.org/wp-content/uploads/2023/06/IDTA-01001-3-0_SpecificationAssetAdministrationShell_Part1_Metamodel.pdf#Page=36) mentions an equivalence relationship: 

> (Submodel)https://example.com/aas/1/1/1234859590
> is identical to
> (GlobalReference)https://example.com/aas/1/1/1234859590

At first glance, this is logical in the sense that both References refer to the same (Submodel) object. 

However, there is a difference between `ModelReference` and `GlobalReference`, as `ModelReference`s are resolvable in implementations of the AAS (since it references another AAS object), whereas `GlobalReference` does not necessarily refer to other AAS objects. 
Furthermore, in the case of `AssetAdministrationShell/submodel`, the spec requires a `ModelReference[Submodel]` anyway, making this equivalence relationship confusing and outdated, as far as I can see. 

This issue came up due to an AAS wrongly modelled under this assumption in [basyx-python-sdk#231](https://github.com/eclipse-basyx/basyx-python-sdk/issues/231)

BirgitBoss:
closes #350 Update clause on matching algorithms
for References
closes #347 Matching with referredSemanticId

Adoption of terms and definition to 
IEC 63278-1:2023
changed terms are listed in change log
abbreviations also adopted if used in 
IEC 632789-1

Terms and definitions were based on a
draft version from July 2022 before

fix Annex Templates UML (<>)

additionally formatting issues fixed for tables 
in "Common Attributes" and "General" chapter

editorial


documentation/IDTA-01001/modules/ROOT/pages/IDTA-01001_General.adoc
```diff
@@ -445,57 +475,91 @@ Note: if the key types are not identical although all key values follow the corr
...
 The definition of link:https://www.w3.org/TR/xmlschema-2/#terminology[XML Schema] is used for matching
 
-"
+
 * (Of string or names:) Two strings or names being compared must be identical.
+
 * Characters with multiple possible representations in ISO/IEC 10646 (e.g. characters 
 with both precomposed and base+diacritic forms) match only if they have the same representation 
-in both strings.
+in both strings
```
s-heppner, line 483: Why has the `.` been removed?

BirgitBoss:
@JoergNeidig

g1zzm0 (approved):
lgtm

JoergNeidig (approved):
Approve. Thank you for the update.

Issue #330, opened 2024-01-06
[BUG] missing rdf prefix in SHACL's SPARQL queries 


mhrimaz:
**Describe the bug**
SHACL shape uses SPARQL queries. However, some queries use rdf namespace. When you try to load the SHAPE with a tool like Apache Jena, it can't parse the queries due to missing `rdf` namespace.

Maybe discussions here would help:

https://groups.google.com/g/topbraid-users/c/Pdb30wZtPlo

**Where**
SHACL File -> SPARQL constraints with rdf. 

https://github.com/admin-shell-io/aas-specs/blob/f4662a61cb7cdb180a779da38b409e7ead749628/schemas/rdf/shacl-schema.ttl#L730-L740

**Screenshots**
![image](https://github.com/admin-shell-io/aas-specs/assets/17963017/022520de-1686-43df-b5b0-8e9d002f36f5)

First option: use sh:declare
![image](https://github.com/admin-shell-io/aas-specs/assets/17963017/8a9bb704-a3d2-4397-abb5-5487a3efcd80)

Second option: use prefix in sparql query
![image](https://github.com/admin-shell-io/aas-specs/assets/17963017/ba7ef62b-4a90-475e-be50-4f5b946d90ec)

@sebbader

PR #356, opened 2024-02-18
Fix SHACL

mhrimaz:
Fixes https://github.com/admin-shell-io/aas-specs/issues/330 and few other issues.


mristin:
@mhrimaz thanks for the fix -- but since the schemas are automatically generated by aas-core-codegen, your changes will be silently overwritten.

Can you please first sync up with @sebbader, and once you confirmed the changes propagate them to:
https://github.com/aas-core-works/aas-core-codegen/blob/main/aas_core_codegen/rdf_shacl/shacl.py

mhrimaz:
Either SHACL or official example should be adapted regarding the `AbstractLangString` in IEC related attributes.
image

And this is the issue with regex format:
image

mhrimaz:
Will be resolved in https://github.com/aas-core-works/aas-core-codegen/pull/526

PR #359, opened 2024-02-22
Provide schemas with unconstrained PathType

s-heppner:
These schemata are not tested yet and are meant
for reference only.
They are not the offical final version.

Since the PathType does not follow RFC 8089
anymore, but the current implementation of the
xs:anyURI regex produces non-compliant XSD,
as a temporary fix to allow compliant XSD schema
generation, we remove the invariant completely.

**This PR is for reference only and should be 
closed, once #375 is fixed!**


BirgitBoss:
I see a line "  `$id": "https://admin-shell.io/aas/3/1",` at the beginning of the xml file. What exactly is is it meant for? Does it denote the version of the specification (this would be IDTA-01001-3-1)? Or the prefix of the elements in the metamodel?
If the latter: since V3.1 we have individual versioning for single classes: not all of them are V3.1.

PR #375, opened 2024-03-07
Rework Chapter "Value-Only serialization" + editorial changes

BirgitBoss:
Complete rework of chapter "Value-Only" (IDTA-01001_ValueOnly.adoc)

Examples added for $value payloads other than submodel to make it easy to understand after example for payload of element within Submodel or structured element.

Aditionally editorial changes were made w.r.t. linking (xref).

### DIFF
--- a/documentation/IDTA-01001/modules/ROOT/pages/IDTA-01001_ChangeLog.adoc
+++ b/documentation/IDTA-01001/modules/ROOT/pages/IDTA-01001_ChangeLog.adoc
@@ -46,7 +46,6 @@ Major Changes:
 
 
 * New value "Role" to enumeration AssetKind (https://github.com/admin-shell-io/aas-specs/issues/294[#294])
-
 * Change PathType: support complete URI scheme for files RFC2396 conformant to anyURI of XML Schema 1.0, Part 2 (https://github.com/admin-shell-io/aas-specs/issues/299[#299])
 * Data type  Identifier: change length from 2000 to 2024 characters (https://github.com/admin-shell-io/aas-specs/issues/306[#306])
 * Referable/idShort and Constraint AASd-002: now also allows hyphens to be included in name (https://github.com/admin-shell-io/aas-specs/issues/295[#295])
@@ -75,7 +74,8 @@ also abbreviations partly adopted; changes:
 (https://github.com/admin-shell-io/aas-specs/issues/350[#350])
 	** introduce equivalent matching and rename exact matching to value matching
 	** added notes 
-* improve ValueOnly documentation for elements other than Submodel (https://github.com/admin-shell-io/aas-specs/issues/371[#371], https://github.com/admin-shell-io/aas-specs/issues/370[#370])
+* transfer of chapters on formats Metadata, Paths and Value-Only from Part 2 API to Part 1 Metamodel (https://github.com/admin-shell-io/aas-specs/issues/325[#325])
+* introduce grammar for Value-Only serialization and enhance documentation (https://github.com/admin-shell-io/aas-specs/issues/371[#371], https://github.com/admin-shell-io/aas-specs/issues/370[#370])
 * Transfer from .docx to asciidoc (.adoc) and maintenance in github
 
 
--- a/documentation/IDTA-01001/modules/ROOT/pages/IDTA-01001_ChangeLog.adoc
+++ b/documentation/IDTA-01001/modules/ROOT/pages/IDTA-01001_ChangeLog.adoc
@@ -75,6 +75,7 @@ also abbreviations partly adopted; changes:
 (https://github.com/admin-shell-io/aas-specs/issues/350[#350])
 	** introduce equivalent matching and rename exact matching to value matching
 	** added notes 
+* improve ValueOnly documentation for elements other than Submodel (https://github.com/admin-shell-io/aas-specs/issues/371[#371], https://github.com/admin-shell-io/aas-specs/issues/370[#370])
 * Transfer from .docx to asciidoc (.adoc) and maintenance in github
 
 
--- a/documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
+++ b/documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
@@ -425,7 +425,7 @@ For a _MultiLanguageProperty_ the Value-Only payload is minimized to the followi
 [
     {"de": "Das ist ein deutscher Bezeichner"},
     {"en": "That's an English label"}
-  ]
+]
 
 ----
 
--- a/documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
+++ b/documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
@@ -79,6 +79,10 @@ There is no Value-Only serialization for an Operation. However, a Value-Only ser
 |RelationshipElement a|first, second
 |AnnotatedRelationshipElement a|first, second, annotations
 
+====
+The Value-Only serialization of the annotations is created the same way as the Value-Only serialization of a SubmodelElementCollection.
+====
+
 2+|*DataElements*
 |Property a| (value)
 
--- a/documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
+++ b/documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
@@ -11,55 +11,221 @@ Plattform Industrie 4.0; Anna Salari, Publik. Agentur für Kommunikation GmbH, d
 ////
 
 
-=== Format "Value" (ValueOnly-Serialization) in JSON
+=== Format "Value" (Value-Only Serialization) in JSON
 
 
-In many cases, applications using data from Asset Administration Shells already know the Submodel regarding its structure, attributes, and semantics. Consequently, there is not always a need to receive the entire model information, which can be requested separately via _Content_ modifier set to _Metadata_, in each request since it is constant most of the time. Instead, applications are most likely only interested in the values of the modelled data. Furthermore, having limited processing power or limited bandwidth, one use case of this SerializationModifier is to transfer data as efficiently as possible. Semantics and data might be split into two separate architecture building blocks. For example, a database would suit the needs for querying semantics, while a device would only provide the data at runtime. Two separate requests make it possible to build up a user interface (UI) and show new upcoming values highly efficiently.
+In many cases, applications using data from Asset Administration Shells already know the Submodel regarding its structure, attributes, and semantics, for example via the used Submodel Template. Consequently, there is not always a need to receive the entire model information. Instead, applications are most likely only interested in the values of the modelled data. Furthermore, having limited processing power or limited bandwidth, one use case of the Format "Value" is to transfer data as efficiently as possible. Semantics and data might be split into two separate architecture building blocks. For example, a database would suit the needs for querying semantics, while a device would only provide the data at runtime. Two separate requests make it possible to build up a user interface (UI) and show new upcoming values highly efficiently.
 
 Values are only available for
 
 * All subtypes of abstract type _DataElement_,
-* SubmodelElementList and SubmodelElementCollection resp. for their included SubmodelElements,
+* SubmodelElementList and SubmodelElementCollection, resp. and for their included SubmodelElements,
 * ReferenceElement,
 * RelationshipElement + AnnotatedRelationshipElement,
 * Entity,
 * BasicEventElement,
 * Submodel.
 
-Capabilities are excluded from the SerializationModifier’s scope since only data containing elements are in the focus. They are consequently omitted in the serialization.
+Capabilities are excluded from the Value-Only serialization scope since only data containing elements are in the focus. They are consequently omitted in the serialization.
 
-The following rules shall be adhered to when serializing a submodel with the SerializationModifier _Value_:
+The Value-Only Format does not contain Metadata attributes (see xref:IDTA-01001_Mappings.adoc#table:meta-data-attributes[Metadata Attributes]). So it is quite the other way around:
 
-* A submodel is serialized as an unnamed JSON object.
-* A submodel element is considered a leaf submodel element if it does not contain other submodel elements. A leaf submodel element follows the rules for the different submodel elements considered in the serialization, as described below. If it is not a leaf element, the serialization rules must be transitively followed until the value is a leaf submodel element.
-* For each submodel element:
-** _Property_ is serialized as $\{Property/idShort}: $\{Property/value} where $\{Property/value} is the JSON serialization of the respective property’s value in accordance with the data type to value mapping (see table after this section).
-** _MultiLanguageProperty_ is serialized as named JSON object with $\{MultiLanguageProperty/idShort} as the name of the containing JSON property. The JSON object contains an array of JSON objects for each language of the _MultiLanguageProperty_ with the language as name and the corresponding localized string as value of the respective JSON property. The language name is defined as two chars according to ISO 639-1.
-** _Range_ is serialized as named JSON object with $\{Range/idShort} as the name of the containing JSON property. The JSON object contains two JSON properties. The first is named “min”. The second is named “max”. Their corresponding values are $\{Range/min} and $\{Range/max}.
-** _File_ and _Blob_ are serialized as named JSON objects with $\{File/idShort} or $\{Blob/idShort}as the name of the containing JSON property. The JSON object contains two JSON properties. The first refers to the content type named $\{File/contentType} resp. $\{Blob/contentType}. The latter refers to the value named “value” $\{File/value} resp. $\{Blob/value}. The resulting ValueOnly object is indistinguishable whether it contains File or Blob attributes. Therefore, the receiver needs to take the type of the target resource into account. Since the receiver knows in advance if a File or a Blob SubmodelElement shall be manipulated, it can parse the transferred ValueOnly object accordingly as a File or Blob object.
-** _SubmodelElementCollection_ is serialized as named JSON object with $\{SubmodelElementCollection/idShort} as the name of the containing JSON property. The elements contained within the struct are serialized according to their respective type with $\{SubmodelElement/idShort} as the name of the containing JSON property.
-** _SubmodelElementList_ is serialized as a named JSON array with $\{SubmodelElementList/idShort} as the name of the containing JSON property. The elements in the JSON array are the ValueOnly serializations of the elements contained in the SubmodelElementList while preserving the order, i.e. index n in the JSON array is the ValueOnly serialization of the element at index n of the SubmodelElementList.
-** _ReferenceElement_ is serialized as $\{ReferenceElement/idShort}: $\{ReferenceElement/value} where $\{ReferenceElement/value} is the serialization of the _Reference_ class.
-** _RelationshipElement_ is serialized as named JSON object with $\{RelationshipElement/idShort} as the name of the containing JSON property. The JSON object contains two JSON properties. The first is named “first”. The second is named “second”. Their corresponding values are $\{RelationshipElement/first} resp. $\{Relationship/second}. The values are serialized according to the serialization of a _ReferenceElement_ (see above).
-** _AnnotatedRelationshipElement_ is serialized according to the serialization of a _ReleationshipElement_ (see above). Additionally, a third named JSON object is introduced with “annotations” as the name of the containing JSON property. The value is $\{AnnotatedRelationshipElement/annotations}. The values of the array items are serialized depending on the type of the annotation data element.
-** _Entity_ is serialized as named JSON object with $\{Entity/idShort} as the name of the containing JSON property. The JSON object contains three JSON properties. The first is named “statements” $\{Entity/statements} and contains an array of the serialized submodel elements according to their respective serialization mentioned in this clause. The second is named either “globalAssetId” or “specificAssetId” and contains either a _Reference_ (see above) or a _SpecificAssetId_. The third property is named “entityType” and contains a string representation of $\{Entity/entityType}.
-** _BasicEventElement_ is serialized as named JSON object with $\{BasicEventElement/idShort} as the name of the containing JSON property. The JSON object contains one JSON property named “observed” with the corresponding value of $\{BasicEventElement/observed} as the standard serialization of the _Reference_ class.
-** _SpecificAssetId_ is serialized as named JSON object with three JSON properties named as the attributes of _SpecificAssetId._
+[[table:value-data-attributes]]
+.Metadata Attributes
+[%autowidth, width="100%", cols="48%,52%",options="header",]
+|===
+|*Class Name* |*Fields for Value-Only representationfootnote:[(value) means that the attribute name is not relevant, only its value, 
+for the other attributes the attribute itself will be seen in the payload]*
+2+|*Identifiables*
+|AssetAdministrationShell |assetInformation, submodels
+|Submodel |submodelElements
+2+|*SubmodelElements*
+|SubmodelElementCollection | (value)
+|SubmodelElementList | (value)
+|Entity |statements, globalAssetId, specificAssetId
+|BasicEventElement |observed
+|Capability |--
+|Operation |--
+2+|*DataElements*
+|Property | (value)
+|MultilanguageProperty | (value)
+|Range |min, max
+|ReferenceElement |value
+|RelationshipElement |first, second
+|AnnotatedRelationshipElement |first, second, annotations
+|Blob |value, contentType
+|File |value, contentType
+|===
 
-* Submodel elements defined in the submodel other than the ones mentioned above are not subject to serialization of that SerializationModifier.
-
-
-
-The following rules shall be adhered to when serializing a submodel element different than a submodel with the SerializationModifier _Value_:
-
-* The submodel element is serialized as an unnamed JSON object.
-* A submodel element is considered a leaf submodel element if it does not contain other submodel elements. A leaf submodel element follows the rules for the different submodel elements considered in the serialization, as described for a submodel above. If it is not a leaf element, the serialization rules must be transitively followed until the value is a leaf submodel element.
+Exception:
+Property/valueId and MultiLanguagePropertyId are also no Metadata attributes. However, they are ignored in the Value-Only serialization.
 
 A submodel element is omitted if no value is defined for it for the requesting submodel or submodel element.
 
-*Data type to value mapping* footnote:[cf. https://openmanufacturingplatform.github.io/sds-documentation/bamm-specification/2.0.0/datatypes.html]
+==== Grammar Value-Only Serialization
 
-The serialization of submodel element values is described in the following table. The left column “Data Type” shows the data types which can be used for submodel element values. The data types are defined according to the W3C XML Schema (https://www.w3.org/TR/xmlschema-2/#built-in-datatypes and https://www.w3.org/TR/xmlschema-2/#built-in-derived). “Value Range” further explains the possible range of data values for this data type. The right column comprises related examples of the serialization of submodel element values.
+In the following the grammar for Value-Only serialization is defined in a formal way:
+
+[listing]
+....
+
+<value-only-of-Submodel> ::= 
+    <value-only-of-SubmodelElementCollection>
+
+<value-only-of-sme> ::= 
+    <value-only-of-SubmodelElementCollection> | <value-only-of-SubmodelElementList> | 
+    <value-only-of-DataElement> | <value-only-of-RelationshipElement> | 
+    <value-only-of-AnnotatedRelationshipElement> | <value-only-of-Entity> | 
+    <value-only-of-BasicEvent>
+
+<value-only-of-DataElement> ::=  
+    <value-only-of-Property> | <value-only-of-MultiLanguageProperty> | 
+    <value-of-Range> | <value-of-ReferenceElement> |  
+    <value-only-of-File> | <value-only-of-Blob> 
+
+
+<value-only-of-SubmodelElementCollection> ::=
+     '{' 
+        {<jd><idShort-of-sme><jd>':' <jd><value-only-of-sme><jd>','}*  
+         <jd><idShort-of-sme><jd>':' <jd><value-only-of-sme><jd>
+        |
+        ''
+     '}'
+
+<value-only-of-SubmodelElementList> ::= 
+    '[' 
+       {<value-only-of-sme>','}* 
+	    <value-only-of-sme> | 
+       ''
+    ']' 
+    
+<value-only-of-Property> ::= 
+	<value-of-Property> 
+
+<value-only-of-MultiLanguageProperty> ::=
+   '['
+      { '{' <language-tag>':' <value-string> '},' }* 
+        '{' <language-tag>':' <value-string> '}'
+   ']'
+
+<value-only-of-Range> ::=  
+    '{  
+		"min":' <value of Range>',' 
+       '"max":' <value of Range>
+    '}'
+
+<value-of-ReferenceElement> ::= 
+    <value-Reference>
+
+<value-of-File> ::= 
+    '{
+        "Contenttype":' <value-ContentType>','
+       '"value":' <value-PathType>
+    '}'
+
+<value-of-Blob> ::= 
+    '{'
+       '"ContentType":' <value-ContentType>[','
+       '"value":' <value-BlobType> ]
+    '}'
+
+<value-only-of-RelationshipElement> ::= 
+    '{
+        "first":' <value-Reference>','
+       '"second":' <value-Reference> 
+    '}'
+    
+<value-only-of-AnnotatedRelationshipElement> ::=
+    '{
+        "first"': <value-Reference>','
+       '"second"': <value-Reference>[','
+       '"annotations": ['  
+            {<jd><idShort-of-sme><jd>':' <value-only-of-DataElement>','}*  
+             <jd><idShort-of-DataElement><jd>':'  <value-only-of-sme>
+        ']'
+    '}'
+        
+<value-only-of-Entity> ::=
+    '{'
+      ['"statements": ['
+          {<jd><idShort-of-sme><jd>':' <value-only-of-sme>','}*  
+           <jd><idShort-of-sme><jd>':' <value-only-of-sme>
+        '],']
+       '"entityType":' <value-EntityType>[',' (
+       '"globalAssetId":' <value-Identifier>[','
+       '"specificAssetIds":' <value-of-specificAssetIds>','] | 
+       '"specificAssetIds":' <value-of-specificAssetIds> )
+       ]
+    '}'
+    
+<value-only-of-BasicEvent> ::= 
+    '{
+        "observed":' <value-Reference>
+    '}'
+    
+<value-of-Property> ::= 
+	value of attribute Property/value with data type as specified in Property/valueType. 
+	In case of data type "xs:string" delimiters '"' are added (e.g. "this-is-mystring")
+
+<value-of-Range> ::= 
+	value of Range/min or Range/max, resp. 
+	with data type as specified in Range/valueType
+
+<value-Reference> ::= 
+    '{'
+       '"type":' <value-of-Reference-type>','
+       '"keys":' <value-of-Reference-keys>
+    '}'
+
+<idShort-of-sme> ::= 
+	value of SubmodelElement/idShort
+
+<language-tag> ::= 
+	language tag as defined for values of type langString, i.e.
+	in accordance with IETF https://tools.ietf.org/html/rfc5646#page-5[RFC 5646], the language names match the following regular expression:
+
+		^[a-z]\{2,4}(-[A-Z][a-z]\{3})?(-([A-Z]\{2}|[0-9]\{3}))?$
+	
+	e.g. "en" for Englisch
+
+<value-of-Reference-type> ::= 
+	value of Reference/type
+
+<value-of-Reference-keys> ::= 
+	"Normal" JSON serialization of list of elements of type Key
+
+<value-of-specificAssetIds> ::= 
+	"Normal" JSON serialization of list of elements of type SpecificAssetId
+
+<value-string> ::= 
+	value of type string with delimiters '"', e.g. "this is my string"
+
+<value-PathType> ::= 
+	value of type PathType
+
+<value-BlobType> ::= 
+	value of type BlobType
+
+<value-Identifier> ::= 
+	value of type Identifier
+
+<jd> ::= 
+	<json-delimiter>  
+
+<sd> ::= 
+	<string-delimiter>  
+
+<json-delimiter> ::= 
+	'"'
+
+<string-delimiter> ::= 
+	'"'
+....
+
+
+==== Data Type to Value Mapping
+
+The serialization of submodel element values is described in the following tablefootnote:[cf. https://eclipse-esmf.github.io/samm-specification/2.0.0/datatypes.html]. The left column “Data Type” shows the data types which can be used for submodel element values. The data types are defined according to the W3C XML Schema 1.0 (https://www.w3.org/TR/xmlschema-2/#built-in-datatypes and https://www.w3.org/TR/xmlschema-2/#built-in-derived). “Value Range” further explains the possible range of data values for this data type. The right column comprises related examples of the serialization of submodel element values.
 
 .Mapping of Data Types in ValueOnly-Serialization
 [%autowidth, width="100%", cols="15%,15%,9%,30%,31%",options="header",]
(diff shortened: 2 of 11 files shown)


documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
```diff
@@ -48,6 +48,15 @@ The following rules shall be adhered to when serializing a submodel with the Ser
 
 * Submodel elements defined in the submodel other than the ones mentioned above are not subject to serialization of that SerializationModifier.
 
+
+
+The following rules shall be adhered to when serializing a submodel element different than a submodel with the SerializationModifier _Value_:
+
+* The submodel element is serialized as an unnamed JSON object.
```
mjacoby, line 55: This statement is wrong, as to my understanding, a property should be serialized only as its value, e.g. `"some string property value"`. This is obviously not a JSON object but still valid JSON (see https://www.json.org/)

documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
```diff
@@ -48,6 +48,15 @@ The following rules shall be adhered to when serializing a submodel with the Ser
...
+* A submodel element is considered a leaf submodel element if it does not contain other submodel elements. A leaf submodel element follows the rules for the different submodel elements considered in the serialization, as described for a submodel above. If it is not a leaf element, the serialization rules must be transitively followed until the value is a leaf submodel element.
```
mjacoby, line 56: > A leaf submodel element follows the rules for the different submodel elements considered in the serialization, as described for a submodel above
This is also not correct, as the section above defined a property to be serialized as `${Property/idShort}: ${Property/value}` but in fact it should just `${Property/value}`.

I think we are making things much more complicated than they are by differentiating between leaf and non-leaf elements. What is wrong about just making a "recursive" definition, e.g. 

Property is serialized as 
> ${Property/Value}

SubmodelElementCollection is serialized as
> {
   "${SubmodelElement/idShort}": valueOnly(${SubmodelElement})    // for each submodel element in the collection
}

etc?

This would cover all cases at once and reduce the complexity of this whole section drastically.

documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
```diff
@@ -48,6 +48,15 @@ The following rules shall be adhered to when serializing a submodel with the Ser
...
+
+Optional submodel elements are omitted if not specified for the requesting submodel or submodel element.
```
mjacoby, line 58: What does that even mean? Either they are part of the serialization or not. What would be an example for this rule to apply that is not covered by other rules/definitions?

documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
```diff
@@ -148,7 +157,7 @@ Full serialization of single submodel element _Property_:
 }
 ----
 
-With the SerializationModifier set to _Value,_ the payload is minimized to the following:
+With the SerializationModifier set to _Value,_ the payload within a submodel or structured submodel element is minimized to the following:
```
mjacoby, line 160: What is a **structured** submodel element? Is it something with children, i.e. a non-leaf element?

documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
```diff
@@ -157,10 +166,16 @@ With the SerializationModifier set to _Value,_ the payload is minimized to the f
...
+[source,json,linenums]
+----
+{
+  5000
+}
+----
 
 
-For a _SubmodelElementCollection,_ the struct is serialized as objects denoted by curly brackets:
+For a _SubmodelElementCollection_, the struct is serialized within a submodel or structured submodel element as objects denoted by curly brackets:
```
mjacoby, line 178: again, what is a **structured** submodel element?

documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
```diff
@@ -186,7 +214,20 @@ For a _SubmodelElementList,_ the struct is serialized as array denoted by square
 }
 ----
 
-For a _MultiLanguageProperty_ named “Label”, the payload is minimized to the following:
+If the _SubmodelElementList_ itself is the subject for ValueOnly the serialization is minimized to the following:
+
+[source,json,linenums]
+----
+ [
```
mjacoby, line 221: FYI, previously it stated that everything is serialized as a JSON object - this is not a JSON object but a JSON array. They are considered different things in JSON.

documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
```diff
@@ -260,7 +323,23 @@ For the same _ReferenceElement,_ the payload is minimized to the following in ca
 }
 ----
 
-For a _File_ named “Document”, the payload is minimized to the following:
+If the _Reference_ itself is the subject for ValueOnly the serialization is minimized to the following:
+[source,json,linenums]
+
+[source,json,linenums]
```
mjacoby, line 329: duplicate of line 327? I am not an expert on AsciiDoc but it looks like an error.

documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
```diff
@@ -260,7 +323,23 @@ For the same _ReferenceElement,_ the payload is minimized to the following in ca
 }
 ----
 
-For a _File_ named “Document”, the payload is minimized to the following:
+If the _Reference_ itself is the subject for ValueOnly the serialization is minimized to the following:
```
mjacoby, line 326: This comment is actually on unchanged line 313, the expected result for serializing a ReferenceElement.

This serialization is wrong as it still contains the idShort of the element, while it should only contain the value, i.e. I would expect

````JSON
{
    "type": "ExternalReference",
    "keys": [
        {
            "type": "GlobalReference",
            "value": "0173-1#02-BAA120#008"
        }
    ]
}
```

documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
```diff
@@ -294,8 +385,18 @@ If the SerializationModifier Extent is set to *_WithBlobValue_*, there is an add
   }
 }
 ----
+ 
+If the _Blob_ itself is the subject for ValueOnly the serialization for Extent set to *_WithBlobValue_  is minimized to the following:
```
mjacoby, line 389: What is the result if extent is set to WithoutBlobValue?

documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
```diff
@@ -372,7 +507,46 @@ For an _AnnotatedRelationshipElement_ named “CurrentFlowFrom”, with an annot
...
+          "type": "Submodel",
+          "value": "http://customer.com/demo/aas/1/0/1234859123490"
+        },
+        {
+          "type": "Property",
+          "value": "MinusPole"
+        }
+      ]
+    },
+    "annotation": [
```
mjacoby, line 541: This maybe needs an explanation...I that the annotation points to a SubmodelElementCollection that, in-turn, contains a single Property of type String with value "TechnicalCurrentFlowDirection". 
Not sure if the original model, i.e. not serialized with content=value but with content=normal is in the document so the reader can understand this.

documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
```diff
@@ -395,7 +569,28 @@ For an _Entity_ named “MySubAssetEntity”, the payload is minimized to the fo
...
-For a BasicEventElement named “MyBasicEvent”, the payload is minimized to the following:
+If the _Entity_ itself is the subject for ValueOnly the serialization is minimized to the following:
+
+[source,json,linenums]
+----
+{
+    "statements": {
+      "MaxRotationSpeed": 5000
+    },
+    "entityType": "SelfManagedEntity",
```
mjacoby, line 580: Is this really part of what is considered the **value** of an Entity? I would expect it to be considered metadata are therefore excluded from valueOnly serialization.

documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
```diff
@@ -418,6 +613,27 @@ For a BasicEventElement named “MyBasicEvent”, the payload is minimized to th
 }
 ----
 
+If the _BasicEventElement_ itself is the subject for ValueOnly the serialization is minimized to the following:
+
+[source,json,linenums]
+----
+{
+    "observed": {
```
mjacoby, line 621: Is *observed* the only property that is considered **value** of a BasicEventElement? If so, why do we need that extra JSON object to wrap around and not simplify the serialization to
```JSON
{
     "type": "ModelReference",
      "keys": [
        {
          "type": "Submodel",
          "value": "http://customer.com/demo/aas/1/1/1234859590"
        },
        {
          "type": "Property",
          "value": "CurrentValue"
        }
      ]
    }
```

mjacoby (changes requested):
There are some things I consider wrong/incorrect.

Even more importantly, I think we are making things much more complicated than they are by differentiating between leaf and non-leaf elements. What is wrong about just making a "recursive" definition, similar to a grammar, e.g.

Property is serialized as

${Property/Value}

SubmodelElementCollection is serialized as

{
"${SubmodelElement/idShort}": valueOnly(${SubmodelElement}) // for each submodel element in the collection
}

etc?

This would cover all cases at once and reduce the complexity of this whole section drastically.

documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
BirgitBoss: one that is no leaf element, might be inconsistent

documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
BirgitBoss: If we would extend it by other (optional) value-attributes then the serialization would not be valid any longer

documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
```diff
@@ -186,7 +214,20 @@ For a _SubmodelElementList,_ the struct is serialized as array denoted by square
...
+If the _SubmodelElementList_ itself is the subject for ValueOnly the serialization is minimized to the following:
+
+[source,json,linenums]
+----
+[
+    "Martha ExampleFamily",
+    "Jonathan ExampleFamily",
+    "Clark ExampleFamily"
+]
+----
```
BirgitBoss, line 226: This seems not to be correct since e.g. collections within a list do not have idShorts but this would be requested here.

documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
BirgitBoss, line 541:
```suggestion
    "annotations": [
```

documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
BirgitBoss, line 58:
```suggestion
A submodel element is omitted if no value is defined for it for the requesting submodel or submodel element.
```

BirgitBoss:
@mjacoby after discussion in TF API I completely reworked the chapter. Please check again: also your findings should be covered now. Thank you

documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
BirgitBoss: It would not be backward-compatible to change this. So it needs to stay like this. If you want to change it please raise an issue for R4.0.

BirgitBoss:
The schema needs to be checked against the openAPI specification. @sebbader : can we just include it as is?

The followings bugs as far as I can see:
* it needs to be "annotations", not "annoation" 
* Blob/value is not required 
* EntityValue: statements are not required, entityType is not required any longer
* although making sense the metamodel itself does not require Range to be numbers 

There are also elements in the schema that are only defined in Part 2, e.g. OperationResultValueOnly

The grammar could be adapted to use the same terms as the schema

documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
```diff
@@ -11,46 +11,221 @@ Plattform Industrie 4.0; Anna Salari, Publik. Agentur für Kommunikation GmbH, d
...
+<value-Identifier> ::= 
+	value of type Identifier
+
+<jd> ::= 
+	<json-delimiter>  
+
+<sd> ::= 
+	<string-delimiter>  
+
+<json-delimiter> ::=
```
mjacoby, line 218: Conflate with `<jd>` definition - only have one of both and do not "chain" them together. This just complicates things more than necessary. Maybe even remove them completely and use literals `'"'` instead

documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
```diff
@@ -100,13 +275,13 @@ Note: the examples are written in RDF/Turtle syntax, and only "Hello" and "Hallo
...
+Note 1: due to the limits in the representation of numbers in JSON, the maximum integer number that can be used without losing precision is 2⁵³-1 (defined as Number.MAX_SAFE_INTEGER). Even if the used data type would allow higher or lower values, they cannot be used if they cannot be represented in JSON. Affected data types are unbounded numeric types xs:decimal, xs:integer, xs:positiveInteger, xs:nonNegativeInteger, xs:negativeInteger, xs:nonPositiveInteger and the bounded type xs:unsignedLong. Other numeric types are not affected. footnote:[cf. https://eclipse-esmf.github.io/samm-specification/2.0.0/datatypes.html (with adjustments for +/-INF, NaN, and language-typed literal support)]
```
mjacoby, line 284: I am not an expert, but where is `Number.MAX_SAFE_NUMBER` defined? This seems to be implementation-specific, i.e. depend on the runtime/framework/programming language used. JSON itself does not define any boundaries for numbers. Please provide are clearer explanation and/or a source.

documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
```diff
@@ -410,15 +607,14 @@ For a BasicEventElement named “MyBasicEvent”, the payload is minimized to th
...
+          "value": "MaxRotation"
         }
       ]
     }
-  }
 }
 ----
 
-===== JSON-Schema for the ValueOnly-Serialization
+==== JSON-Schema for the ValueOnly-Serialization
```
mjacoby, line 617: The JSON Schema (which was not changed, i.e. I cannot make a comment on that during review) includes elements that are not defined in the meta model/part 1, namely `OperationRequestValueOnly` and `OperationResultValueOnly`.
These should not be part of this document but there should be a section in part 2 adding to this JSON schema for classes added in part 2 meta model.

mjacoby (changes requested):
Most important issues are

- using a grammar requires properties in JSON to have exact same ordering as defined in grammar, but that is not what we want as ordering of properties in a JSON object is arbitrary
- Some elements are currently not allowed to be empty, e.g. AnnotatedRelationshipElement/annotations but are allowed to be empty according to meta model
- mixed use of literal `'"'` and `<jd>` in grammar - use only one of those, preferable the literal

In general, I'm not sure using a formal grammar is the best way to represent this as it brings other side effects such as ordering, additional white spaces, line breaks, etc.
I suppose it would be clearer and easier to read to provide a textual description along with some JSON-style template of how the payload should look like and some examples.

documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
```diff
@@ -157,10 +166,16 @@ With the SerializationModifier set to _Value,_ the payload is minimized to the f
 }
 ----
 
-TODO: Example where precision in value string to ValueOnly to value string is critical.
+If the property itself is the subject for ValueOnly the serialization is minimized to the following:
```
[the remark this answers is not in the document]
BirgitBoss: True, will be corrected

documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
BirgitBoss: sorry, a structured element do have child elements. SMC and SML

BirgitBoss: reformulated, structured submodel element not used any longer in text

documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
```diff
@@ -11,46 +11,221 @@ Plattform Industrie 4.0; Anna Salari, Publik. Agentur für Kommunikation GmbH, d
...
+        {<jd><idShort-of-sme><jd>':' <value-only-of-sme>','}*  
+         <jd><idShort-of-sme><jd>':' <value-only-of-sme>
+        |
+        ''
+     '}'
+
+<value-only-of-SubmodelElementList> ::= 
+    '[' 
+       {<value-only-of-sme>','}* 
+	    <value-only-of-sme> |
```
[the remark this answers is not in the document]
BirgitBoss: ok

documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
BirgitBoss: removed

documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
BirgitBoss: true

documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
```diff
@@ -11,46 +11,329 @@ Plattform Industrie 4.0; Anna Salari, Publik. Agentur für Kommunikation GmbH, d
...
+    "NameOfFather": "Jonathan ExampleFamily",
+    "NameOfSon": "Clark ExampleFamily"
+}
+----
+    
+====
+
+* Submodel is an exception. Its attribute "submodelElements" is treated like the attribute "value" of a submodel element collection.
+
+* For complex types the serialization is identical to the "Normal"-format serialization. If this is not the case its serialization needs to be explicitly defined.
```
mjacoby, line 169: What exactly is a `complex type`?

documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
```diff
@@ -11,46 +11,329 @@ Plattform Industrie 4.0; Anna Salari, Publik. Agentur für Kommunikation GmbH, d
...
+	e.g. "en" for Englisch
+|===
 
-* Submodel elements defined in the submodel other than the ones mentioned above are not subject to serialization of that SerializationModifier.
 
-*Data type to value mapping* footnote:[cf. https://openmanufacturingplatform.github.io/sds-documentation/bamm-specification/2.0.0/datatypes.html]
+A submodel element is omitted if no value is defined for it for the requesting submodel or submodel element.
 
-The serialization of submodel element values is described in the following table. The left column “Data Type” shows the data types which can be used for submodel element values. The data types are defined according to the W3C XML Schema (https://www.w3.org/TR/xmlschema-2/#built-in-datatypes and https://www.w3.org/TR/xmlschema-2/#built-in-derived). “Value Range” further explains the possible range of data values for this data type. The right column comprises related examples of the serialization of submodel element values.
+==== How-To Value-Only Serialization
```
mjacoby, line 131: This definition is better than the previous one but I have the feeling this could still be improved.
Maybe it could make sense to introduce different types of element, e.g.  (rules apply recursively)

- primitive: will be directly serialized as value
   - `Property`
- complex: will be serialized using default serialization
   - `Reference` (actually, I am not sure if this is not correct because the valueOnly serialization of a reference probably does not include e.g. referredSemanticId but this is actually undefined right now I would say)
   - `MultiLanguageProperty`?
   - `Blob`?
   - `File`?
- composite: serialized as JSON object using properties defined in above table in the format "idshort": {value}
   - `Submodel`
   - `SubmodelElementCollection`
   - `Entity`
   - ...
- SubmodelElementList (seems to be a special case): serialized as JSON array 


Das ist definitiv noch nicht ausgereift sondern nur ein erster Gedanke von mir. Wäre interessant zu schauen ob man das zu einer brauchbaren Definition aufbohren kann, dafür habe ich allerdings nicht die Ressourcen.

documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
```diff
@@ -11,46 +11,329 @@ Plattform Industrie 4.0; Anna Salari, Publik. Agentur für Kommunikation GmbH, d
...
+	
+....	
+
+	e.g. "en" for Englisch
+|===
 
-* Submodel elements defined in the submodel other than the ones mentioned above are not subject to serialization of that SerializationModifier.
 
-*Data type to value mapping* footnote:[cf. https://openmanufacturingplatform.github.io/sds-documentation/bamm-specification/2.0.0/datatypes.html]
+A submodel element is omitted if no value is defined for it for the requesting submodel or submodel element.
```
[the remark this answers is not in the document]
BirgitBoss: I do expect "tons of empty elements" for optional elements. This is why they are optional. As far as I remember there was a decision to remove them from the payload. And the sentence shall reflect this decision.l

documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
```diff
@@ -11,46 +11,329 @@ Plattform Industrie 4.0; Anna Salari, Publik. Agentur für Kommunikation GmbH, d
...
+|File a|value, contentType
+
+2+|*Data Types*
+|OperationVariable a| (value)
+
+|MultiLanguageTextType a| The value of this type is identicial to its "Normal"-format serialization: it consists of a list of pairs: a language tag and the value in the corresponding language
+
+[listing]
+....
+<language-tag> ::=
```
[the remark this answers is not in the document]
BirgitBoss: this is true but it can be repeated here ?

documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
BirgitBoss: see [comment ](https://github.com/admin-shell-io/aas-specs/pull/375#discussion_r1551038092)above: you require definitions for "complex type" etc. but here you also do not provide definitions, just a (complete) list of elements that needs to be maintained.

documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
```diff
@@ -11,46 +11,329 @@ Plattform Industrie 4.0; Anna Salari, Publik. Agentur für Kommunikation GmbH, d
...
+|Submodel a| (submodelElements)
+
+====
+Qualifiers are ignored. This also holds for qualifiers of kind "ValueQualifier".
+====
+
+|*SubmodelElements* a| 
+
+====
+Qualifiers are ignored. This also holds for qualifiers of kind "ValueQualifier".
```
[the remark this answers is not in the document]
BirgitBoss: no, the note above was for submodels, this is for submodel elements

documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
```diff
@@ -11,46 +11,329 @@ Plattform Industrie 4.0; Anna Salari, Publik. Agentur für Kommunikation GmbH, d
...
+[%autowidth, width="100%", cols="48%,52%",options="header",]
+|===
+|*Class Name* |*Fields for Value-Only representationfootnote:[ attribute in parenthesis (like e.g. (value)) means that the attribute name is not displayed, only its Value-Only serialization. 
+For the other attributes the attribute itself will be part of the serialization]*
+2+|*Identifiables*
+|AssetAdministrationShell a|assetInformation, submodels
+|Submodel a| (submodelElements)
+
+====
+Qualifiers are ignored. This also holds for qualifiers of kind "ValueQualifier".
```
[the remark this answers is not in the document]
BirgitBoss: This is true

documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
```diff
@@ -11,46 +11,329 @@ Plattform Industrie 4.0; Anna Salari, Publik. Agentur für Kommunikation GmbH, d
...
-** _BasicEventElement_ is serialized as named JSON object with $\{BasicEventElement/idShort} as the name of the containing JSON property. The JSON object contains one JSON property named “observed” with the corresponding value of $\{BasicEventElement/observed} as the standard serialization of the _Reference_ class.
-** _SpecificAssetId_ is serialized as named JSON object with three JSON properties named as the attributes of _SpecificAssetId._
+[[table-value-data-attributes]]
+.Value-Only Attributes
+[%autowidth, width="100%", cols="48%,52%",options="header",]
+|===
+|*Class Name* |*Fields for Value-Only representationfootnote:[ attribute in parenthesis (like e.g. (value)) means that the attribute name is not displayed, only its Value-Only serialization. 
+For the other attributes the attribute itself will be part of the serialization]*
+2+|*Identifiables*
+|AssetAdministrationShell a|assetInformation, submodels
```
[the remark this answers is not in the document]
BirgitBoss: removed attributes and added note that there is no serialization. If needed we can introduce it later.

Why assetInformation used in valueOnly : because the rest is Metainformation. We cannot change it because of backward compatibility. I added footnote for it.

documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
BirgitBoss: Link from ESMF: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER added + link corrected

documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
```diff
@@ -11,46 +11,329 @@ Plattform Industrie 4.0; Anna Salari, Publik. Agentur für Kommunikation GmbH, d
...
 * RelationshipElement + AnnotatedRelationshipElement,
 * Entity,
 * BasicEventElement,
 * Submodel.
 
-Capabilities are excluded from the SerializationModifier’s scope since only data containing elements are in the focus. They are consequently omitted in the serialization.
+Capabilities are excluded from the Value-Only serialization scope since only data containing elements are in the focus. They are consequently omitted in the serialization.
 
-The following rules shall be adhered to when serializing a submodel with the SerializationModifier _Value_:
+The Value-Only Format does not contain Metadata attributes (see xref:IDTA-01001_Mappings.adoc#format-metadata-metadata-serialization[Metadata Attributes]). So the serialization contains the value-related data, only and ignored the metadata:
```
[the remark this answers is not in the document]
BirgitBoss: ok

documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
```diff
@@ -11,46 +11,329 @@ Plattform Industrie 4.0; Anna Salari, Publik. Agentur für Kommunikation GmbH, d
...
+
+====
+Qualifiers are ignored. This also holds for qualifiers of kind "ValueQualifier".
+====
+
+|SubmodelElementCollection a| (value) 
+
+====
+If the collection contains submodel elements of type
+ Operation or Capability: these are ignored in the Value-Only serialization
```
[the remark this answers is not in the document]
BirgitBoss: I am not sure: Operation and capability do not have a Value-Only serialization but this could also mean that if a collection contains them the complete collection is not serialized. But here it is stated that they are ignored and the rest is serialized. I think this needs to be clarified.

documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
```diff
@@ -11,46 +11,329 @@ Plattform Industrie 4.0; Anna Salari, Publik. Agentur für Kommunikation GmbH, d
...
+
+====
+valueId is ignored in Value-Only serialization
+====
+|Range a|min, max
+
+type equal to Range/valueType
+
+====
+In the Value-Only serialization only number ranges are supported
```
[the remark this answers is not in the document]
BirgitBoss: It was specified like this in the previsous version in Part 2.
For me also Ranges only make sense with Numbers. But in Part 1 we allow any xsd type from DataTypeDefXsd.
It would be backward compatible to change this in the Value-Only Schema. We should discuss it.

BirgitBoss:
@mjacoby please review again, table was completely revised and rules adapted (and corrected)

documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
mjacoby: The provided link does not back what is claimed here. First, it has no connection at all to JSON; the word JSON is not even mentioned in the linked document. Second, it explicitly says in the first sentence, that this only applies to JavaScript.
Please add a source (or at least detailed explanation) for the following statement 
> due to the limits in the representation of numbers in JSON, the maximum integer number that can be used without losing precision is 2⁵³-1 

I am not against having this statement per-se, but as I never have heard of this and currently don't understand why this would be the case I think this requires proof.

documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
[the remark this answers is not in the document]
mjacoby: It seems it did mix up things and you are correct, assetInformation is **not** part of metadata serialization and therefore there is no contradiction in having it as part of valueOnly serialization.

However, I still recommend removing this section completely if it is not used because it is just cluttering the document without bringing any benefits. If kept, I propose to move assetInformation to the metadata serialization format instead of valueOnly as I would argue this data is rather considered metadata than a value.

In the end, if you really insist on keeping things as-are, I could also live with that as it has no actual effect because it is never used.

documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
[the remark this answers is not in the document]
mjacoby: Indeed, it needs to be defined what is the expected result when a collection is empty or only contains elements that are not serializable. 
My suggestion would be that they could be omitted, i.e. empty elements are optional.

However, duplicating the information that Operation and Capability are not serializable for valueOnly is not really helpful. I suggest changing the test to something similar to
> If a SubmodelElementCollection is empty or contains only SubmodelElements that have no valueOnly serialization, serializing the SubmodelElementCollection may be omitted.

documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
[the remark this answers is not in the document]
mjacoby: We discussed this issue yesterday in the API workstream and decided Part 1 is correct and valueOnly serialization should be updated accordingly.

documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
mjacoby: You are right, the proposed changes are not helping. 
My idea was to make this table shorter by not repeating elements where regular serialization and valueOnly serialization are the same, e.g. Blob, File, Reference(?). Only the types in the category I called `composite` need to be explained further.

The current proposal just adds complexity without real benefit so I am against it. Either keep it as it was previously, i.e. a table of all element types listed explicitly or try to go the way I proposed and find a way to shorten that table by re-using previous definitions (e.g. default serialization for some elements)

documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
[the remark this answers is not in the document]
mjacoby: It seems like how to handle empty elements is one of the most important question. This especially becomes important when thinking about serialization of SubmodelElementLists.

What is the valueOnly serialization of a SubmodelElementList containing 2 properties, first one is empty, second one is not. Omitting empty values this would result in 
```JSON
[
   "some value"
]
```

and it would be impossible to tell if `some value` is the value of the first or second element in the list.

Omitting every empty element could lead to somewhat strange result, e.g. a submodel containing multiple collections, each only containing elements with empty value would just be serialized to an completely empty string following this logic. Not sure if that is what I would expect.

Furthermore, I do not have the feeling that having "tons of empty elements" would be a realistic problem because there are also parameters like level which allows to limit the depth of the returned structure. 

Not sure what is the best solution here but just enforcing all empty elements to be omitted obviously causes problems.

mjacoby:
@BirgitBoss I've responded to the changes/comments

documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
BirgitBoss: Sorry, this was your proposal to go. Sounded more easy than it is. If you have something different in mind then please  provide a proposal for the simplified table.

documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
BirgitBoss: not in text any longer

documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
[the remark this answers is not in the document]
BirgitBoss: ok, change needs to be done

documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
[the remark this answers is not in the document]
BirgitBoss: I added issue https://github.com/admin-shell-io/aas-specs-api/issues/270 for this discussion.

The discussion with the empty collection seems to be the same as in https://github.com/admin-shell-io/aas-specs/pull/375#discussion_r1558986741

I think a collection cannot be empty in the sense that a value is missing. It is valid that are not SubmodelElements in the collection but I would display it  as an empty object because this is what was defined by the data provider. In the Normal-Serialization "value" could be ommitted but not the collection itself. It would be part of the Value-Only serialization of the Submodel. So it is different to Capability or Operation that are always "empty" w.r.t. this understanding.

documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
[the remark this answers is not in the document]
BirgitBoss: I would not agree that assetInformation is metadata in the senses of $metadata: alle data in AAS is metadata in some sense but some are metadata for the asset being represented (assetInformation) and some for the AAS as object created by a data provider etc. (like e.g. versioning).

BirgitBoss:
> @BirgitBoss I've responded to the changes/comments

Thanks a lot, I added my responses.

documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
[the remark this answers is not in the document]
BirgitBoss: I added issue https://github.com/admin-shell-io/aas-specs/issues/398 to request 

`For of Property within a Submodel with kind=Instance Property/value shall be defined.`

`For of MultiLanguageProperty within a Submodel with kind=Instance MultiLanguageProperty/value shall be defined.`

documentation/IDTA-01001/modules/ROOT/pages/IDTA-01001_ChangeLog.adoc
```diff
@@ -75,6 +74,8 @@ also abbreviations partly adopted; changes:
 (https://github.com/admin-shell-io/aas-specs/issues/350[#350])
 	** introduce equivalent matching and rename exact matching to value matching
 	** added notes 
+* transfer of chapters on formats Metadata, Paths and Value-Only from Part 2 API to Part 1 Metamodel (https://github.com/admin-shell-io/aas-specs/issues/325[#325])
+* introduce grammar for Value-Only serialization and enhance documentation (https://github.com/admin-shell-io/aas-specs/issues/371[#371], https://github.com/admin-shell-io/aas-specs/issues/370[#370])
```
BirgitBoss, line 78:
```suggestion
* enhance documentation for Value-Only serialization (https://github.com/admin-shell-io/aas-specs/issues/371[#371], https://github.com/admin-shell-io/aas-specs/issues/370[#370])
```

documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
mjacoby: My proposal was not to just add subheadings to the table without any kind of explanation what that mean or why they are there. I already provided different categories than you used and also included description/reasoning for each category.

documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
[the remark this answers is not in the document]
mjacoby: I get your point. However, if the idea is to divide all properties into either metadata or value, I would consider `assetInformation` to rather be metadata than value.

documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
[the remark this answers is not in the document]
mjacoby: I agree that there might be some misunderstanding regarding the term *optional* because it may refer to the meta-model or to JSON. My primary issue is that we agree on a JSON serialization and once we have that we can write the rules how to generate it.

Regarding Property valueOnly serialization, what is the correct/expected JSON output for serializing a submodel (idShort *submodel*) that contains a SubmodelElementCollection (idShort *collection*) which again contains a single Property with idShort *property* and no value?

A simplified version of regular JSON serialization could look like this
```JSON
{
	"modelType": "Submodel",
	"idShort": "submodel",	
	"submodelElements": [
		{
			"modelType": "SubmodelElementCollection"
			"idShort": "collection",
			"value": [
				{
					"modelType": "Property",
					"idShort": "property",
					"valueType": "xs:string"					
				}
			]
		}
	]	
}
```

I can think of the following possible results

1. 
```JSON
{
    "collection": {
        "property": null
    }
}
```
2. 
```JSON
{
    "collection" : { }
}
```
3. 
```JSON
{ }
```
4. 
```JSON
null
```

At first glance, only the first option looks viable to me because of the issue with SumodelElementLists as discussed in an earlier comment.
Using this to deduct serialization rules from it, this would imply that all values have to be serialized at all times resp. `null` to be used to indicate a value is not present.

documentation/IDTA-01001/modules/ROOT/pages/includes/IDTA-01001_ValueOnly.adoc
BirgitBoss: But the example states that the annotation is a Property, not a Collection... the property is not part of a collection. But you are right: annotations is treated like SubmodelElementCollection w.r.t. its serialization. This text could be added.

BirgitBoss:
https://github.com/admin-shell-io/aas-specs/tree/SeBa/part-2-bugfix-changes is already closed. So branch https://github.com/admin-shell-io/aas-specs/tree/BiBo/ValueOnly should be merged in IDTA-01001_working directly.

PR #377, opened 2024-03-08
Close #376 semanticId ExternalReferences and Close #366 individual versioning

### LINKED ISSUE #366: Individual versioning of metamodel element IDs
Up to V3.0 all elements got a new version, e.g. `https://admin-shell.io/aas/3/1/AssetAdministrationShell` even if there was no change.

If there are no changes then the version should not be increased. Inherited class changes need to be considered in the version. So if the class A that is inherited by class B is changed then also the version of class B is increased.
Same for type changes. If a type A is changed and used as type in one of the attributes in class B then the version of class B needs to be increased as well.
The third case is the change of constraints: if a constraint has an impact on the allowed values of a class B the the version of class B needs to be increased.

Same for API operations in Part 2

### LINKED ISSUE #376: HasSemantics/semanticId shall always be an External Reference

image

BirgitBoss:
#376 Improvement of note recommending to use external references for HasSemantics/semanticId

#366 reverse individual versioning

### DIFF
--- a/documentation/IDTA-01001/modules/ROOT/pages/Spec/IDTA-01001_Metamodel_Common.adoc
+++ b/documentation/IDTA-01001/modules/ROOT/pages/Spec/IDTA-01001_Metamodel_Common.adoc
@@ -39,7 +39,7 @@ Other attributes of the administrative information like creator refer to ISO 158
 For more information on the concept of subject, see Attribute Based Access Control (ABAC) link:#bib49[[49\]]. The assumption is that every subject has a unique identifier.
 
 
-xref:AdministrativeInformation[AdministrativeInformation] allows the usage of templates (xref:HasDataSpecification[HasDataSpecification]). Data specifications are defined in separate documents.
+xref:Spec/IDTA-01001_Metamodel_Common.adoc#AdministrativeInformation[AdministrativeInformation] allows the usage of templates (xref:Spec/IDTA-01001_Metamodel_Common.adoc#HasDataSpecification[HasDataSpecification]). Data specifications are defined in separate documents.
 
 
 ====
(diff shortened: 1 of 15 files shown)


mhrimaz:
@BirgitBoss I am very sorry that I ask my question here. 
If in the future we makes it mandatory that semanticId should be `ExternalReference/GlobalReference`, then why we even have `ModelReference/ConceptDescription`?
@s-heppner This will be closely related to https://github.com/eclipse-basyx/basyx-python-sdk/issues/234 and https://github.com/eclipse-basyx/basyx-python-sdk/issues/244.

In the end, how can I understand that a semanticId like `https://www.hsu-hh.de/aut/aas/physicaladdress` is just a url or a `ConceptDescription` that I can find in some ConceptDescriptionRepository?

BirgitBoss:
There are other attributes with type "Reference" that are not affected. 
In general a data consumer does not know the endpoints where to get concept descriptions. Additional discovery services are needed. This is especially true in decentralized systems.

documentation/IDTA-01001/modules/ROOT/pages/Spec/IDTA-01001_Metamodel_Common.adoc
```diff
@@ -303,7 +303,10 @@ Identifier of a supplemental semantic definition of the element called supplemen
 
 
 ====
-Note: it is recommended to use an external reference.
+Note: it is recommended to use an external reference. This is even recommended in the case a xref:IDTA-01001_Metamodel_Designators.adoc#ConceptDescription[ConceptDescription] exists. Otherwise lookup of the corresponding concept description may fail.
+
+In future releases a constraint may be added that makes is mandatory to use external references.
```
s-heppner, line 308:
```suggestion
In future releases a constraint may be added that makes it mandatory to use external references.
```

Small typo

s-heppner: What is implied here, but not directly stated, is that we do not want the user to reference a "local" ConceptDescription that for example is shipped with an AASX package, right?

I fear that this formulation will create more AASX packages where constraint AASd-123 is violated: 

> Constraint AASd-123: For model references, i.e. References with Reference/type = ModelReference, the value of Key/type of the first key of Reference/keys shall be one of AasIdentifiables.

The reason for this is that someone who manually creates an AAS sees that one should refer to a `ConceptDescription` with an `ExternalReference` and do so, even if the referenced `ConceptDescription` is located inside the model and therefore needs a `ModelReference`.

s-heppner:
What is implied here, but not directly stated, is that we do not want the user to reference a "local" ConceptDescription that for example is shipped with an AASX package, right?

I fear that this formulation will create more AASX packages where constraint AASd-123 is violated: 

> Constraint AASd-123: For model references, i.e. References with Reference/type = ModelReference, the value of Key/type of the first key of Reference/keys shall be one of AasIdentifiables.

The reason for this is that someone who manually creates an AAS sees that one should refer to a `ConceptDescription` with an `ExternalReference` and do so, even if the referenced `ConceptDescription` is located inside the model and therefore needs a `ModelReference`.

BirgitBoss:
> The reason for this is that someone who manually creates an AAS sees that one should refer to a `ConceptDescription` with an `ExternalReference` and do so, even if the referenced `ConceptDescription` is located inside the model and therefore needs a `ModelReference`.

But we do not state that a concept description shall be referenced. we state that is shall be a concept definition. This is the superterm for any kind of concept definition including concept descriptions which are model elements.

s-heppner:
I get that, however I still think the formulation may lead to confusion, with users potentially utilizing external references to reference model objects. 

Maybe we could clarify the note text with something like this?

```
Note: it is recommended to use an external reference, (e.g. referencing a globally resolvable semantic definition). This...
```

BirgitBoss:
@s-heppner please check again: I incorportated your change proposals

BirgitBoss:
@mjacoby could you also review (added notes in line 309 and line 294, the other changes are just xref corrections)

mjacoby:
@BirgitBoss I do not feel qualified to review this as I have neither participated in this discussion nor am I familiar with AsciiDoc and those xrefs.

BirgitBoss:
> @BirgitBoss I do not feel qualified to review this as I have neither participated in this discussion nor am I familiar with AsciiDoc and those xrefs.

sure, xref need to review, only the note. But I will remove you from the reviewer list, no problem

PR #378, opened 2024-03-08
Change /3/0 to /3/1 semanticId for MM elements

### LINKED ISSUE #376: HasSemantics/semanticId shall always be an External Reference

BirgitBoss:
Decison to use indivual versioning for classes and attributes was 
revised. Therefore change  /3/0 to /3/1 semanticId for metamodel 
elements

+ upate changelog
Close #376


g1zzm0 (approved):
LGTM

Issue #401, opened 2024-04-11
Improve documentation of Value-Only


BirgitBoss:
**Is your feature request related to a problem? Please describe.**

The existing rules are hard to understand. Many readers requested a more easy to understand documentation of the Value-Only serialialiation.

Additionally, Examples shall be updated to not come from Submodel but provide Value-only for every element since the API (IDTA-01002) allows $value calls not only for submodels.

see also 
https://github.com/admin-shell-io/aas-specs-api/issues/270
https://github.com/admin-shell-io/aas-specs-api/issues/269
https://github.com/admin-shell-io/aas-specs-api/issues/268

BirgitBoss:
@mjacoby listed the remaining remaining problems in
https://github.com/admin-shell-io/aas-specs/pull/403#pullrequestreview-1996355586
 
- [x] **(1)** 2^53-1 is again missing the 5
- [x] **(2)** JSON schema again contains validation for Operations which are defined not to have a valueOnly serialization
- [ ] **(3)** Range is again limited to number
- [ ] **(4)** There are no rules defined how to serialize elements other than `Submodel`, i.e. serialization of all other types is only given by a single example. This means serialization is seriously underdefined and unclear. 
- [ ] **(5)** The section about limitation of numbers in JSON to 2^53-1 is still lacking any substantial reasoning. There is no such limitation in JSON itself, only in the implementations in different programming languages - and they might even be different to each other. So why this limitation?
- [ ] **(6)** Handling of non-existent values is not addressed at all, i.e. what is the valueOnly Serialization of a property that has no value? What about SubmodelElementCollection without elements or Submodel without any elements?
- [ ] **(7)** Missing serialization rules for auxiliary classes (i.e. classes that are not subclass of SubmodelElement but still part of the serialization) like `Reference`
- [ ] **(8)** Rules for serialization in section `Rules for Value-Only Serialization` are inconsistent/overlapping, e.g., 
  The definition of `SubmodelElementCollection` says
  > SubmodelElementCollection is serialized as named JSON object with $\{SubmodelElementCollection/idShort} as the name of the containing JSON property. The elements contained within the struct are serialized according to their respective type with $\{SubmodelElement/idShort} as the name of the containing JSON property.

  while the definition of `Property` says
  > Property is serialized as $\{Property/idShort}: $\{Property/value} where $\{Property/value} is the JSON serialization of the respective property’s value in accordance with the data type to value mapping (see table after this section).

[...]

PR #372, opened 2024-03-04
Add changes for idShortPath and ValueOnly fixes from part 2 v3.0.2

sebbader-sap:
Adaptation of adoc should complie to latest [valueOnly](https://rwth-iat.github.io/aas-specs-antora/IDTA-01001/snapshot/IDTA-01001_Mappings.html#_format_value_valueonly_serialization_in_json)and [IdShortPath ](https://rwth-iat.github.io/aas-specs-antora/IDTA-01001/snapshot/IDTA-01001_Mappings.html#_format_path_idshortpath_serialization_in_json)Specification

PR #403, opened 2024-04-11
#401 Update Chapter on Value-Only Serialization - first changes

BirgitBoss:
Chapter on Value-Only Serialization was updated:
- update examples to provide Value-Only serialization 
        (not in the context of a Submodel)
- add table similar to metadata table: which attributes are displayed
- add chapter for Submodel Example 
      (with properties from SMT Technical Data, not Families any longer)
- bugfix annotations instead of annoation in schema and examples
- editorial bugfixes: update deprecated links

part of #401


BirgitBoss:
@mjacoby @sebbader-sap As discussed in todays Workstream AAS Meeting: here are the changes agreed to (see description of PR). 

For any improvements of the chapter on the rules etc. a separate PR would be needed: this just shows first improvements.

P.S. https://github.com/admin-shell-io/aas-specs/pull/372 from @sebbader w.r.t. changes of V3.0.2 of Part 1 was already accepted before

mjacoby (changes requested):
I am ok with the changes made but there are multiple issues that need to be addressed before considering this issue to be completed. As they reference unchanged code, I cannot comment directly inline. Instead, I will make a list with checkboxes so that we can keep track of what is still to do and put numbers in there so you can reference a specific issue.
It also seems that some errors fixed and issues discussed in previous PRs (e.g. https://github.com/admin-shell-io/aas-specs/pull/375) are back. 

- [x] **(1)** 2^53-1 is again missing the 5
- [x] **(2)** JSON schema again contains validation for Operations which are defined not to have a valueOnly serialization
- [ ] **(3)** Range is again limited to number
- [ ] **(4)** There are no rules defined how to serialize elements other than `Submodel`, i.e. serialization of all other types is only given by a single example. This means serialization is seriously underdefined and unclear. 
- [ ] **(5)** The section about limitation of numbers in JSON to 2^53-1 is still lacking any substantial reasoning. There is no such limitation in JSON itself, only in the implementations in different programming languages - and they might even be different to each other. So why this limitation?
- [ ] **(6)** Handling of non-existent values is not addressed at all, i.e. what is the valueOnly Serialization of a property that has no value? What about SubmodelElementCollection without elements or Submodel without any elements?
- [ ] **(7)** Missing serialization rules for auxiliary classes (i.e. classes that are not subclass of SubmodelElement but still part of the serialization) like `Reference`
- [ ] **(8)** Rules for serialization in section `Rules for Value-Only Serialization` are inconsistent/overlapping, e.g., 
  The definition of `SubmodelElementCollection` says
  > SubmodelElementCollection is serialized as named JSON object with $\{SubmodelElementCollection/idShort} as the name of the containing JSON property. The elements contained within the struct are serialized according to their respective type with $\{SubmodelElement/idShort} as the name of the containing JSON property.

  while the definition of `Property` says
  > Property is serialized as $\{Property/idShort}: $\{Property/value} where $\{Property/value} is the JSON serialization of the respective property’s value in accordance with the data type to value mapping (see table after this section).

  If you apply these rules recursively like required by another rule, you end up with something containing the `idShort` of the property twice.

BirgitBoss:
I added an updated ChangeLog for the changes done

BirgitBoss:
I will merge this PR, For additional findings please see https://github.com/admin-shell-io/aas-specs/pull/408

PR #404, opened 2024-04-11
Fix refs and remove bugs

zrgt:
- Removed "page$" from xrefs
- Removed docu.html file
- Removed bugs 
  - While generating the .adoc file from word file
    some bug texts were created, probably from word metadata. 
In this PR these bugs were fixed.


BirgitBoss:
Why is it preferrable to not use page$ in xref?

zrgt:
> Why is it preferrable to not use page$ in xref?

Because such links do not work in generated antora website.

Issue #412, opened 2024-04-16
Optional values for all SubmodelElements


alexgordtop:
The constraints / cardinalities affecting the value of the SubmodelElements are inconsistent.

It is not obvious, why especially ReferenceElement.value, but also Property.value, File.value, Blob.value are optional, but e.g. RelationshipElement.first/second is required.

**Request:**
Please revise required attributes with the aspect of defining a template structure / incomplete model in mind.

----

**Issue caused by restriction:**
This mandatory constraint leads currently to issues with AASX serialization, since some IDTA SubmodelTemplates would like to specify the usage of e.g. RelationshipElements, but when deserializing a corresponding AASX, the parsers will fail according to a xml schema violation.

---

Beside SubmodelTemplates, there are other UseCases, where a kind of requested structure should be provided, that somebody else would have to fill like: 

**Use case:**
A component has an empty configuration, that the operator should fill to configure the asset.
Using RelationshipElements is nearly impossible in the situation.

The same applies to File/Blob.contentType - if a component allows an attachment, it is forced to specify something like "*/*", if it is open for different kinds of values. From a modeler perspective, I would have preferred to not set the contentType in advance and describe in the description, that a kind of image or pdf is accepted as documentation, but no videos.

BirgitBoss:
TF Metamodel 2024-04-17
Discussion not yet finalized but recommendation:
- at least for (Annotated)RelationshipElement make "first" and "second" optional (conformant to min and max of Range)

BirgitBoss:
Decision Proposal TF Part 1 2024-05-15
Besides https://github.com/admin-shell-io/aas-specs/issues/412#issuecomment-2061377393 make the following attributes optional as well:
**File/contentType
Blob/contentType**

Reason:

- contentType may not be fixed at time of creation of a SMT
- even for a Submodel with kind=Instances representing a configuration file that is sent to another stakeholder for filling in the gaps the information might not be available

BirgitBoss:
Workstream AAS Sepc 2024-06-13
accepted
make the following attributes optional

- (Annotated)RelationshipElement/first 
- (Annotated)RelationshipElement/second
- File/contentType
- Blob/contentType

PR #416, opened 2024-04-18
work on #412: make Relationship/first and /second optional

BirgitBoss:
#412 make Relationsship/first and Relationsship/second optional
+ update figures affected
+ update ChangeLog
+ add note that most value-related attributes optional because of SMT


alexgordtop (approved):
Besides commit message length seems to be ok

PR #426, opened 2024-05-06
Adapt the schema files to v3.0.1

s-heppner:
This adapts the schema files to version 3.0.1
of the specification.

However, this is not an official release yet
and may be subject to change.
This is only meant as reference for the 
API specification.


schemas/json/aas.json
```diff
@@ -36,7 +36,7 @@
       "properties": {
         "language": {
           "type": "string",
-          "pattern": "^(([a-zA-Z]{2,3}(-[a-zA-Z]{3}(-[a-zA-Z]{3}){2})?|[a-zA-Z]{4}|[a-zA-Z]{5,8})(-[a-zA-Z]{4})?(-([a-zA-Z]{2}|[0-9]{3}))?(-(([a-zA-Z0-9]){5,8}|[0-9]([a-zA-Z0-9]){3}))*(-[0-9A-WY-Za-wy-z](-([a-zA-Z0-9]){2,8})+)*(-[xX](-([a-zA-Z0-9]){1,8})+)?|[xX](-([a-zA-Z0-9]){1,8})+|((en-GB-oed|i-ami|i-bnn|i-default|i-enochian|i-hak|i-klingon|i-lux|i-mingo|i-navajo|i-pwn|i-tao|i-tay|i-tsu|sgn-BE-FR|sgn-BE-NL|sgn-CH-DE)|(art-lojban|cel-gaulish|no-bok|no-nyn|zh-guoyu|zh-hakka|zh-min|zh-min-nan|zh-xiang)))$"
```
sebbader-sap, line 39: > ... |i-klingon| ...

seriously?? :D

schemas/json/aas.json
```diff
@@ -683,7 +683,7 @@
                   "pattern": "^([\\t\\n\\r -\ud7ff\ue000-\ufffd]|\\ud800[\\udc00-\\udfff]|[\\ud801-\\udbfe][\\udc00-\\udfff]|\\udbff[\\udc00-\\udfff])*$"
```
sebbader-sap, line 683:
```suggestion
                  "pattern": "^([\\t\\n\\r-\\ud7ff\\ue000-\\ufffd]|\\ud800[\\udc00-\\udfff]|[\\ud801-\\udbfe][\\udc00-\\udfff]|\\udbff[\\udc00-\\udfff])*$"
```
Also for all other occurrences of this pattern.

sebbader-sap (changes requested):

schemas/json/aas.json
mristin: https://en.wikipedia.org/wiki/Codes_for_constructed_languages#List_of_codes

Incredible, but true ;-)

schemas/json/aas.json
s-heppner: Good eye, I'll adapt this and regenerate the schema files

s-heppner: Actually those appear to be two problems here in one. 
- One is that the Path_type regex is actually still in here, which is relatively easy to fix
- The second one however is this missing backslash. @mristin This appears to be a bug in aas-core-codegen, since the regex looks fine to me in [aas-core-meta](https://github.com/aas-core-works/aas-core-meta/blob/6d5411bf9f5ef0ccc1ed956c63c2eca50711137f/aas_core_meta/v3.py#L365)

mristin: @s-heppner @sebbader-sap I'm working on it today. It's amazing when such strange & obvious bugs slip so many eyes!

mristin: Just for future reference: this is actually not a bug. As long as a character is in the Basic Multilingual Plane (BMP, <0xFFFF), the JSON Schema engines will automatically understand the character. Hence, it does not matter whether you write:
```
"pattern": "\ud7ff"
```

... or:
```
"pattern": "\\ud7ff"
```

The first pattern is given with a character as-is, the second pattern escapes the character. This matters when, say, you print out the violation message (most engines indicate the pattern violates). The first message will contain the pattern with the character as-is and the second message will be shown with the pattern escaped.

This still needs to be fixed for consistency.

sebbader-sap:
@s-heppner what is the state here? Please ping me when I can re-review this PR again.

s-heppner:
@sebbader-sap You can use the now updated version for the API generation, however we still need to run the tests and generate the example files, before I will mark this PR as ready.

schemas/json/aas.json
```diff
@@ -674,18 +674,9 @@
...
...]{1,4}|([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\\.([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\\.([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\\.([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5]))|(([0-9A-Fa-f]{1,4}:){4}[0-9A-Fa-f]{1,4})?::([0-9A-Fa-f]{1,4}:[0-9A-Fa-f]{1,4}|([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\\.([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\\.([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\\.([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5]))|(([0-9A-Fa-f]{1,4}:){5}[0-9A-Fa-f]{1,4})?::[0-9A-Fa-f]{1,4}|(([0-9A-Fa-f]{1,4}:){6}[0-9A-Fa-f]{1,4})?::)|[vV][0-9A-Fa-f]+\\.([a-zA-Z0-9\\-._~]|[!$&'()*+,;=]|:)+)\\]|([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\\.([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\\.([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\\.([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])|([a-zA-Z0-9\\-._~]|%[0-9A-Fa-f][0-9A-Fa-f]|[!$&'()*+,;=])*)))?/((([a-zA-Z0-9\\-._~]|%[0-9A-Fa-f][0-9A-Fa-f]|[!$&'()*+,;=]|[:@]))+(/(([a-zA-Z0-9\\-._~]|%[0-9A-Fa-f][0-9A-Fa-f]|[!$&'()*+,;=]|[:@]))*)*)?|/((([a-zA-Z0-9\\-._~]|%[0-9A-Fa-f][0-9A-Fa-f]|[!$&'()*+,;=]|[:@]))+(/(([a-zA-Z0-9\\-._~]|%[0-9A-Fa-f][0-9A-Fa-f]|[!$&'()*+,;=]|[:@]))*)*)?)$"
```
sebbader-sap, line 686: Shouldn't be there a new pattern with the URI/URL regex?

schemas/json/aas.json
```diff
@@ -1281,18 +1272,9 @@
...
```
sebbader-sap, line 1293: Same as before: Shouldn't this pattern not replaced instead of deleted?

sebbader-sap (changes requested):

schemas/json/aas.json
sebbader-sap: Makes sense

s-heppner:
> https://github.com/admin-shell-io/aas-specs/blob/b0e3e73e0584a869adc62fec30cce94d923eabe6/schemas/json/aas.json#L44
> still has the old regex pattern.

This should be resolved now. Can you please check again?

schemas/xml/AAS.xsd
```diff
@@ -5,7 +5,7 @@
       <xs:element name="language">
         <xs:simpleType>
           <xs:restriction base="xs:string">
-            <xs:pattern value="(([a-zA-Z]{2,3}(-[a-zA-Z]{3}(-[a-zA-Z]{3}){2})?|[a-zA-Z]{4}|[a-zA-Z]{5,8})(-[a-zA-Z]{4})?(-([a-zA-Z]{2}|[0-9]{3}))?(-(([a-zA-Z0-9]){5,8}|[0-9]([a-zA-Z0-9]){3}))*(-[0-9A-WY-Za-wy-z](-([a-zA-Z0-9]){2,8})+)*(-[xX](-([a-zA-Z0-9]){1,8})+)?|[xX](-([a-zA-Z0-9]){1,8})+|((en-GB-oed|i-ami|i-bnn|i-default|i-enochian|i-hak|i-klingon|i-lux|i-mingo|i-navajo|i-pwn|i-tao|i-tay|i-tsu|sgn-BE-FR|sgn-BE-NL|sgn-CH-DE)|(art-lojban|cel-gaulish|no-bok|no-nyn|zh-guoyu|zh-hakka|zh-min|zh-min-nan|zh-xiang)))"/>
+            <xs:pattern value="(([a-zA-Z]{2,3}(-[a-zA-Z]{3}(-[a-zA-Z]{3}){,2})?|[a-zA-Z]{4}|[a-zA-Z]{5,8})(-[a-zA-Z]{4})?(-([a-zA-Z]{2}|[0-9]{3}))?(-(([a-zA-Z0-9]){5,8}|[0-9]([a-zA-Z0-9]){3}))*(-[0-9A-WY-Za-wy-z](-([a-zA-Z0-9]){2,8})+)*(-[xX](-([a-zA-Z0-9]){1,8})+)?|[xX](-([a-zA-Z0-9]){1,8})+|((en-GB-oed|i-ami|i-bnn|i-default|i-enochian|i-hak|i-klingon|i-lux|i-mingo|i-navajo|i-pwn|i-tao|i-tay|i-tsu|sgn-BE-FR|sgn-BE-NL|sgn-CH-DE)|(art-lojban|cel-gaulish|no-bok|no-nyn|zh-guoyu|zh-hakka|zh-min|zh-min-nan|zh-xiang)))"/>
```
sebbader-sap: A new comma is added in the part for `{2}` --> `{,2}`. Isn't the latter one incorrect and should be `{0,2}`?

sebbader-sap:
@s-heppner the current state looks good for me, just two last questions:
1. See my comment above about the cardinality declaration. I am just not sure whether `{,2}` is a valid shortcut for `{0,2}`. Can you confirm this?
2. You removed the spaces from the general String-checking regex pattern, this seems fine therefore. However, I still find patterns with spaces, e.g. https://github.com/admin-shell-io/aas-specs/blob/6ab2e16187c3a3f1a60d722860e553b48ce201e3/schemas/json/aas.json#L692  Is this a problem?

schemas/xml/AAS.xsd
mristin: The zero is optional in regex engines.

Missing zero/comma was a mistake traced back to abnf-to-regexp.

s-heppner:
@sebbader-sap 

ad 1.) The zero is optional in regexes, see [this comment](https://github.com/admin-shell-io/aas-specs/pull/426#discussion_r1633707848)

ad 2.) From all I can tell, this regex is fine and generally the spaces shouldn't place problems. 

Please wait with the release until I was able to run the schema generation one last time, since we found one last issue: [aas-core-works/aas-core-codegen#494](https://github.com/aas-core-works/aas-core-codegen/issues/494)

s-heppner:
I regenerated the JSON Schema and confirmed that there are no other effects of [aas-core-works/aas-core-codegen#494](https://github.com/aas-core-works/aas-core-codegen/issues/494). 

Therefore, this is now ready for release, as far as I'm converned

sebbader-sap (approved):

PR #431, opened 2024-05-15
make contentType of File and Blob optional

BirgitBoss:
Closes https://github.com/admin-shell-io/aas-specs/issues/412 make File/contentType and Blob/contentType optional

+ update ChangeLog
+ update Figures
+ update Value-Only Serialization (also make Range min,max optional as defined in Metamodel)

Additionally, fix DataElement: Stereotype <<abstract>> was missing


BirgitBoss:
may you please review @alexgordtop , thank you

alexgordtop (approved):

Issue #362, opened 2024-02-28
No UTF32 Characters in the Regex for Strings


sebbader-sap:
**Describe the bug**

The regex pattern in the JSON Schema has only UTF-16 characters, while constraint AASd-130 demands the following:
`^[\x09\x0A\x0D\x20-\uD7FF\uE000-\uFFFD\u00010000-\u0010FFFF]*$` 

**Where**
JSON Schema, e.g., https://github.com/admin-shell-io/aas-specs/blob/2ab08f92bdd1d44edc1cfee52552fe5429d2178e/schemas/json/aas.json#L44C22-L44C36

          "pattern": "^([\\t\\n\\r -\ud7ff\ue000-\ufffd]|\\ud800[\\udc00-\\udfff]|[\\ud801-\\udbfe][\\udc00-\\udfff]|\\udbff[\\udc00-\\udfff])*$"

**Additional context**
Needs to be adopted in the SwaggerHub Domains for Part 1 and Part 2.

mristin:
@sebbader-sap we transpiled the patterns into UTF-16 since most JSON schema engines we tested operated on UTF-16 and could not handle UTF-32.

It is a trade-off between correctness and practicality -- if we put UTF-32 in JSON schema (*e.g.*, the pattern you mentioned: `^[\x09\x0A\x0D\x20-\uD7FF\uE000-\uFFFD\u00010000-\u0010FFFF]*$`), many JSON schema validators will fail. I suppose any schema validator written as a library for C# or Java will be in that group.

You can test it online. The first answer on Google for "JSON schema validator" for me is https://www.jsonschemavalidator.net/. This validator does support UTF-32:
```json
{
  "$schema": "https://json-schema.org/draft/2019-09/schema",
  "title": "AssetAdministrationShellEnvironment",
  "type": "string",
  "pattern": "^[\\x09\\x0A\\x0D\\x20-\\uD7FF\\uE000-\\uFFFD\\U00010000-\\U0010FFFF]*$"
}
```

The following value passes:
```
"test"
```

This is not a big change in aas-core-codegen, so whatever the decision, it shouldn't be hard to fix.

mristin:
Please consider also the SHACL -- I think the same issue appears there as well.

sebbader-sap:
The original pattern from the constraint has the same problems with OpenAPI-based validators, as they usually translate the YAML into JSON Schema --> then using the same JSON Schema Validation libraries with the same UTF-32 problems.

[...]

PR #433, opened 2024-05-23
Document pattern deviation from specification

s-heppner:
We document in the JSON and RDF schema `README`
files, that we deviate from the pattern in the
specification of AASd-130, due to the fact that
most schema engines test UTF-16, instead
of the used UTF-32.

For the full discussion, refer to #362

### DIFF
--- a/schemas/json/README.md
+++ b/schemas/json/README.md
@@ -123,6 +123,11 @@ the [mapping of XSD to JSON types] into account.
 
 [5.7.12 Primitive and Simple Data Types]: https://industrialdigitaltwin.org/wp-content/uploads/2023/04/IDTA-01001-3-0_SpecificationAssetAdministrationShell_Part1_Metamodel.pdf#page=96
 
+#### Note: Pattern Deviation from the Specification
+Since most JSON schema engines we tested operated on UTF-16 and could not handle UTF-32, we transpiled the pattern from AASd-130, which uses UTF-32 in the specification, into UTF-16. 
+This is a trade-off between correctness and practicality. 
+See [#362](https://github.com/admin-shell-io/aas-specs/issues/362) for the details.
+
 #### Hint: Round-Trip Conversions
 
 Round-trip conversions XML to JSON to XML or RDF to JSON to RDF may not result in the
--- a/schemas/json/README.md
+++ b/schemas/json/README.md
@@ -124,7 +124,26 @@ the [mapping of XSD to JSON types] into account.
 [5.7.12 Primitive and Simple Data Types]: https://industrialdigitaltwin.org/wp-content/uploads/2023/04/IDTA-01001-3-0_SpecificationAssetAdministrationShell_Part1_Metamodel.pdf#page=96
 
 #### Note: Pattern Deviation from the Specification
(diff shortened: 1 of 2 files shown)


BirgitBoss (approved):

schemas/json/README.md
```diff
@@ -123,6 +123,11 @@ the [mapping of XSD to JSON types] into account.
 
 [5.7.12 Primitive and Simple Data Types]: https://industrialdigitaltwin.org/wp-content/uploads/2023/04/IDTA-01001-3-0_SpecificationAssetAdministrationShell_Part1_Metamodel.pdf#page=96
 
+#### Note: Pattern Deviation from the Specification
+Since most JSON schema engines we tested operated on UTF-16 and could not handle UTF-32, we transpiled the pattern from AASd-130, which uses UTF-32 in the specification, into UTF-16.
```
mristin, line 127:
```suggestion
The patterns given in the specification are written as Unicode character points.
The Unicode character points, which are numbers, can be arbitrarily long.
The Unicode merely regulates how to associate a character point, *i.e.*, a number, with every possibly imaginable character in the world.
There is no prescription how these numbers should be encoded.
The UTF-8 encoding is usually used to encode Unicode character points in files, while many programming languages use UTF-16 and UTF-32 encodings, respectively, to represent strings of Unicode characters in memory.

UTF-32 uses 4 bytes to represents characters.
Since Unicode currently does not define more than 2³² characters, which can all fit into 4 bytes, UTF-32 can represent all Unicode characters as-are directly as numbers.
On the other hand, UTF-16 uses 2 bytes to represent characters, so Unicode character points above 2¹⁶ need to be encoded by two consecutive UTF-16 characters.
Pattern matching operating on UTF-16 needs to take these two consecutive characters into account when character points above 2¹⁶ are to be matched.

There is no unified encoding for JSON Schema engines, so we have to decide which encoding to use.
Most JSON schema engines we tested relied on the programming language they were implemented in for pattern matching, so we assume UTF-16 as the character encoding for Unicode characters.
Notably, C# and Java use UTF-16.

The patterns in the schema need consequently to take into account characters encoded as UTF-16.
To that end, we transpiled the pattern from AASd-130 to match UTF-16-encoded characters.
Hence, the pattern is much longer than in the specification, as Unicode character points above 2¹⁶ need to be matched with two characters in the pattern instead of one.

See [#362](https://github.com/admin-shell-io/aas-specs/issues/362) for the details.
```

schemas/rdf/README.md
```diff
@@ -74,6 +74,11 @@ following:
   validation errors in the shapes. RDF does not contain a concept for abstract classes,
   therefore custom checks using SPARQL queries are supplied.
 
+#### Note: Pattern Deviation from the Specification
+Since most RDFS engines we tested operated on UTF-16 and could not handle UTF-32, we transpiled the pattern from AASd-130, which uses UTF-32 in the specification, into UTF-16. 
+This is a trade-off between correctness and practicality. 
+See [#362](https://github.com/admin-shell-io/aas-specs/issues/362) for the details.
```
mristin, line 80: I don't think that RDFS engines allow for arbitrary test to be encoded. Actually, I imagine that XML strings are used to encode text by default, so AASd-130 is meaningless (all text *must* conform to XML text to be ingested in RDF in the first place).

@sebbader-sap can you please have a look at this point?

mristin (approved):
I wrote a bit longer explanation to avoid future confusion.

PR #464, opened 2024-08-30
Fix bugs in IDTA-01001-3-0-1 schemas

### LINKED ISSUE #457: [BUG] environment tag should be declared as "xs:element" instead of "xs:group" in AAS.xsd
When validating v3.0 AAS XML files against AAS.xsd I get the error that the tag environment is not declared. Digging deeper 
the error is occurring because in the schema file (XSD), the tag "environment" is defined as an "xs:group" instead of "xs:element" which is supposed to be root. The root element of an XML document should always be defined as an "xs:element" in the XSD file like this in our example:
`<xs:element name="environment">`
Changing the XSD accordingly resolves my validation errors. 
I attached the new XSD file (as txt as GitHub does not support xsd) with the ask to adjust the changes in the project.
[AAS_xsd.txt](https://github.com/user-attachments/files/16361537/AAS_xsd.txt)

### LINKED ISSUE #477: Changed Order of `dataSpecificationContent` and `dataSpecification` in XML
**Describe the bug**
[Release v3.0.1](https://github.com/admin-shell-io/aas-specs/commit/87cf203a8c0067cc1315c0fd65c3e1084c11e351#diff-33fa391e7dbbb30b74596de382a4636b499ba32979a3a7e19ebe30afca30626e) has introduced a change also for the attributes `dataSpecificationContent` and `dataSpecification`.

**Where**
The change in order makes every(!) older XML file with embeddedDataSpecifications invalid without any reason.

**Screenshots**
image

**Proposal**
Given that there is no reason for this change, we could either:
1. revert this change, or
2. change the clause in XML that the order for these two attributes shall not be relevant.

s-heppner:
This fixes bugs found in schema files in preparation of a v3.0.9 release of the schema files:

# Fix order of EmbeddedDataSpecification attributes
Previously, the order of the attributes in class
`EmbeddedDataSpecification` got mixed up, making
the schemas of Release 3.0.8 not backward
compatible, as found out in #477.

This reverts the change of the order of attributes
back to the state of v3.0, so that the schemas are
properly backward compatible.

Fixes #477

# SHACL
Make SHACL regex patterns follow JSON Schema ones

We included the regex pattern as-is from the 
input which caused problems with the 
regex engines as the patterns in the meta-model 
are written in a Python dialect 
(and assuming that the regex engine works on 
UTF-32 characters). However, most regex engines 
in the wild operating on SHACL 
(*e.g.*, Java SHACL validators) use UTF-16 
to represent the text and do not support 
some parts of the Python regex dialect. 
For example, in the input meta-model 
specification, we omit the minimum bound 0
(*e.g.*, ``{,4}``), which breaks with the Java 
regex engine beneath the SHACL validator.

Instead, with this patch, we parse the pattern 
from the specification and re-render it into the 
form that we also use in JSON Schema. We pick
JSON Schema regex dialect as most 
SHACL validators in the wild can deal with it, 
in particular those based on Java as a platform. 
Hence, we decide to serve this 
user base with priority.

Discovered in [aas-core-meta issue 342].
Fixed in [aas-core-codegen commit e22cc].

[aas-core-meta issue 342]: aas-core-works/aas-core-meta#342
[aas-core-codegen commit e22cc]: aas-core-works/aas-core-codegen@e22ccae

# XSD
Adapt XML Schema for some schema validators

Previously we did not define `aas:environment`
directly via an `xs:element` tag, rather only 
provided it directly as `xs:complexType`.
This created issues with several XML Schema 
validator engines. 

Fixes #457.

Lastly, we also update the generated examples.


BirgitBoss (approved):

BirgitBoss:
Please update [README.md#releases](https://github.com/admin-shell-io/aas-specs/blob/master/README.md#releases) after making an updated release

BirgitBoss:
see bug issue: https://github.com/admin-shell-io/aas-specs/issues/477

BirgitBoss (changes requested):
https://github.com/admin-shell-io/aas-specs/issues/477 needs to be fixed as well

BirgitBoss:
The `*OverPatternsExamples` seem to be new. May you please explain what they are covering?

BirgitBoss (approved):

PR #465, opened 2024-09-06
#331 update explanation and notes for Administrative Information

### LINKED ISSUE #331: Submodel ID and Administrative Information: too restrict?
Context: Administrative Information Attributes, Notes

```
Note 1: two submodels with the same semanticId but different administrative information shall have different IDs (Submodel/id), since they denote that the submodel is not backward compatible or has some other major administrative changes. The idShort typically does not change. The same applies to other identifiables (Identifiable/id). Otherwise, the ID of a submodel would not be sufficient to identify the data or service provided by the submodel.

Note 2: since submodels with different versions shall have different identifiers, it is possible that an Asset Administration Shell has two submodels with the same semanticId but different versions, i.e. different administrative metainformation.

Note 3: some of the administrative information like the version number might need to be part of the identification. This is similar to the handling of identifiers for concept descriptions using IRDIs. In ECLASS, the IRDI 0173-1#02-AO677#002 contains the version information 002.

Note 4: the process of versioning or adding other administrative information to elements is done by external version or configuration management software and not by the Asset Administration Shell itself.
```

The attribute` AdministrativeInformation/creator `and the `AdministrativeInformation/templateId` was added in V3.0 but the notes were not updated. So now it is requested that if the value of the creator or template ID attribute is changed that a new Submodel ID must be created.

***Proposal:*** Substitute with the following rules:

------
If an AAS contains two different Submodels with the same semanticId (_Submodel/semanticId_) these two submodels shall have different IDs (_Submodel/id_) and differ in either their a) version (_Submodel/administration/version_) (revision is ignored) or b) their creator (_Submodel/administration/creator_). With a) both submodels shall have version information. With b) both submodels shall have a creator.

Note 1: typically, some of the administrative information like the version number might be part of the identification (Submodel/id). This is similar to the handling of identifiers for concept descriptions using IRDIs. In ECLASS, the IRDI 0173-1#02-AO677#002 contains the version information 002.

Note 2: since submodels with different versions shall have different identifiers, it is possible that an Asset Administration Shell has two submodels with the same semanticId but different versions.

If an AAS contains two different Submodels guided by the same SMT, i.e. have the same templateId value (Submodel/administration/templateId), then the two submodels shall have different IDs (Submodel/id). In this case both Submodels shall have a templateId assigned to them (_Submodel/administration/templateId_). 

Note 1: In some cases there is neither a semanticId (_Submodel/semanticId_)  nor a templateId (_Submodel/administration/templateId_) defined for the Submodel. In this case there is no way for the data consumer to formally see whether two Submodels are providing the same semantic information. 

Note 2: If a template ID of one of the standardized Submodel Templates of the IDTA is used to guide the creation of a Submodel  then there is also a semantic ID defined for the Submodel.

------

These rules are backward compatible since they are less restrictive then the current notes.
No "shall-" rules should be contained in notes (conformant to IEC rules of writing standards)

BirgitBoss:
closes #331 update explanation and notes for Administrative Information
+ added [[]] for linking
+ update changelog

### DIFF
--- a/documentation/IDTA-01001/modules/ROOT/pages/spec-metamodel/common.adoc
+++ b/documentation/IDTA-01001/modules/ROOT/pages/spec-metamodel/common.adoc
@@ -69,7 +69,7 @@ Note 3: In some cases there is neither a semanticId (xref:spec-metamodel/common.
 ====
 
 ====
-Note 2: If a template ID of one of the standardized Submodel Templates of the IDTA is used to guide the creation of a Submodel then there is also a semantic ID defined for the Submodel.
+Note 4: If a template ID of one of the standardized Submodel Templates of the IDTA is used to guide the creation of a Submodel then there is also a semantic ID defined for the Submodel.
 ====
 
 
--- a/documentation/IDTA-01001/modules/ROOT/pages/spec-metamodel/common.adoc
+++ b/documentation/IDTA-01001/modules/ROOT/pages/spec-metamodel/common.adoc
@@ -65,7 +65,7 @@ If an AAS contains two different Submodels guided by the same Submodel Template 
 In this case both Submodels shall have a templateId assigned to them (xref:spec-metamodel/common.adoc#templateId[Submodel/administration/templateId]).
 
 ====
-Note 1: In some cases there is neither a semanticId (xref:spec-metamodel/common.adoc#semanticId[Submodel/semanticId]) nor a template ID (xref:spec-metamodel/common.adoc#templateId[Submodel/administration/templateId]) defined for the Submodel. In this case there is no way for the data consumer to formally see whether two Submodels are providing the same semantic information.
+Note 3: In some cases there is neither a semanticId (xref:spec-metamodel/common.adoc#semanticId[Submodel/semanticId]) nor a template ID (xref:spec-metamodel/common.adoc#templateId[Submodel/administration/templateId]) defined for the Submodel. In this case there is no way for the data consumer to formally see whether two Submodels are providing the same semantic information.
 ====
 
 ====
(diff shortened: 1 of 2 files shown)


JoergNeidig (approved):
I agree with the content of the changes. However, before merging the numbering of the "notes" should be checked as there are currently multiple Notes 1 and 2.

documentation/IDTA-01001/modules/ROOT/pages/spec-metamodel/common.adoc
```diff
@@ -46,27 +46,33 @@ The assumption is that every subject has a unique identifier.
...
 
+
+If an AAS contains two different Submodels guided by the same Submodel Template (SMT), i.e. have the same templateId value (Submodel/administration/templateId), then the two submodels shall have different IDs (xref:spec-metamodel/common.adoc#Identifiable[Submodel/id]). 
+In this case both Submodels shall have a templateId assigned to them (xref:spec-metamodel/common.adoc#templateId[Submodel/administration/templateId]).
+
 ====
-Note 3: some of the administrative information like the version number might need to be part of the identification.
-This is similar to the handling of identifiers for concept descriptions using IRDIs.
-In ECLASS, the IRDI 0173-1#02-AO677#002 contains the version information #002.
+Note 1: In some cases there is neither a semanticId (xref:spec-metamodel/common.adoc#semanticId[Submodel/semanticId]) nor a template ID (xref:spec-metamodel/common.adoc#templateId[Submodel/administration/templateId]) defined for the Submodel. In this case there is no way for the data consumer to formally see whether two Submodels are providing the same semantic information.
```
BirgitBoss, line 68:
```suggestion
Note 3: In some cases there is neither a semanticId (xref:spec-metamodel/common.adoc#semanticId[Submodel/semanticId]) nor a template ID (xref:spec-metamodel/common.adoc#templateId[Submodel/administration/templateId]) defined for the Submodel. In this case there is no way for the data consumer to formally see whether two Submodels are providing the same semantic information.
```

documentation/IDTA-01001/modules/ROOT/pages/spec-metamodel/common.adoc
```diff
@@ -46,27 +46,33 @@ The assumption is that every subject has a unique identifier.
...
 ====
 
 ====
-Note 4: the process of versioning or adding other administrative information to elements is done by external version or configuration management software and not by the Asset Administration Shell itself.
+Note 2: If a template ID of one of the standardized Submodel Templates of the IDTA is used to guide the creation of a Submodel then there is also a semantic ID defined for the Submodel.
```
BirgitBoss, line 72:
```suggestion
Note 4: If a template ID of one of the standardized Submodel Templates of the IDTA is used to guide the creation of a Submodel then there is also a semantic ID defined for the Submodel.
```

Issue #385, opened 2024-03-12
[BUG] The spelling of ID-Short-Path should be uniform


g1zzm0:
**Describe the bug**
The spelling of ID-Short-Path should be uniform. In the PDF and ASCII Doc are several spellings of IdShort-Path, IDShortPath, etc.

**Where**
PDF Spec and ASCII Doc

**Screenshots**
![image](https://github.com/admin-shell-io/aas-specs/assets/1814815/241f43eb-4d60-4ed7-9142-e08b1c247921)
![image](https://github.com/admin-shell-io/aas-specs/assets/1814815/9a318b7b-039a-47dc-ae8b-84da5f18a4ca)

**Additional context**
id-short-path could not be found with the Antora search feature.

BirgitBoss:
same for IRDI path and IRDI-Path: use IRDI-Path

zrgt:
Fixed in #469

PR #469, opened 2024-09-19
Implement Minor Fixes and Editorial Changes

zrgt:
- Fix an AAS puml diagramm
- Use "IRDI-Path" instead of "IRDI path". See #385 
- Use "idShortPaths" instead of "idShort paths". See #385 
- Fix some typos and grammar
- Fix unresolved refs
- Add use of PlantUML to changelog
- Remove mentions of Enterprise Architect


BirgitBoss (approved):

BirgitBoss (approved):
Thanks!

The footnote that still mentions Enterprise Architect is perhaps not valid any longer since we use PlantUml

zrgt:
> Thanks!
> 
> The footnote that still mentions Enterprise Architect is perhaps not valid any longer since we use PlantUml

I have removed now the mentions of Enterprise Architect

PR #470, opened 2024-10-05
Update CONTRIBUTING.md

BirgitBoss:
copy version from aas-specs-api with minor editorial changes


BirgitBoss:
@sebbader-sap I think you need to review my changes again: i just committed your proposed changes

sebbader-sap (approved):

PR #482, opened 2024-10-31
fix or improve explanation for References and Semantics

### LINKED ISSUE #473: Update Matching Strategy SemanticId
There are different matching strategies described for semantic IDs. What is not clear from the example is whether the key.type should be considered or not in exact matching etc. (see chapter on reference matching: these two matching strategies need to be combined).

### LINKED ISSUE #479: Add Matching Strategy how to deal with supplementalSemanticID
At the moment it is not described how to handle supplementalSemanticID for the matching of SemanticIDs

### LINKED ISSUE #480: Add Example of Reference String serialization that includes a References with referredSemanticId
At the moment there is no example given how a References with refferedSemanticId is serialized as a string

### LINKED ISSUE #481: [BUG] Example for Reference wrong
**Describe the bug**
[Note ](https://admin-shell-io.github.io/aas-specs-antora/IDTA-01001/v3.1/spec-metamodel/referencing.html#_constraints)

`Note: "(SubmodelElementCollection)0, (MultiLanguageProperty)Title" may be identical to
"(SubmodelElementCollection)Manual, (MultiLanguageProperty)Title" semantically and content-wise. The difference is
that more than one document is allowed in the first submodels and thus a submodel element list is defined:
elements in a list are numbered. However, it is also possible to define a display name in this case. The display
name of the SubmodelElementCollection should be the same in both bases, e.g. "Users Manual".`

is not correct since a number "0" is not allowed as value for idShort

Needs to SubmodelElementList in the first refence. Proposal to remove the hint with the display name since we probably will allow idShort also for Lists.

`Note: "(SubmodelElementList)0, (MultiLanguageProperty)Title" may be identical to
"(SubmodelElementCollection)Manual, (MultiLanguageProperty)Title" semantically and content-wise. The difference is
that more than one document is allowed in the first submodels and thus a submodel element list is defined:
elements in a list are numbered. `

BirgitBoss:
- closes #480 
- closes #481 
- closes #473
- closes #479
- added more invalid reference examples for all constraints effecting References
- fix formatting


BirgitBoss:
@j-blume: may you please also have a look at this PR? Thank you

documentation/IDTA-01001/modules/ROOT/pages/general.adoc
```diff
@@ -446,16 +446,23 @@ When comparing two elements, different use cases should be considered in order t
 This clause gives first hints on the aspects to consider when dealing with matching semantic identifiers.
```
s-heppner, line 446: To someone that only reads this section (as they are only interested in how semantic identifiers are matched at that moment), it might seem like an official server would support all of these strategies. 
I see how this sentence, together with the fact that we're in the non-normative part of the specification should be enough to clarify that this isn't the case, but maybe it might make sense to add something like:

```suggestion
This clause gives first hints on the aspects to consider when dealing with matching semantic identifiers.
The actual implementations in applications may differ from the strategies presented here. 
```

Except if we actually think that "Exact Matching" is normative (considering we actually define equality here), but then it should be clearly stated so.

documentation/IDTA-01001/modules/ROOT/pages/general.adoc
```diff
@@ -444,18 +444,26 @@ This semantic ID helps to easily find the semantic definition of the submodel (s
...
+
 
 *Exact Matching (identical semanticIds) – DEFAULT*
 
-* With exact matching, two semantic IDs need to be string-identical.
-** Example: Property with idShort "ManufacturerName" + semanticId 0173-1#02-AAO677#002 and Property with idShort "Herstellername" + semanticId 0173-1#02-AAO677#002 have exactly equal semantics.
+* With exact matching, two semantic IDs need to represent two exact matching references (see xref:general.adoc#matching-algorithm-for-references[Matching Algorithm for References]).
+** Example: Property with idShort "ManufacturerName" + semanticId  value 0173-1#02-AAO677#002 and Property with idShort "Herstellername" + semanticId value 0173-1#02-AAO677#002 have exactly equal semantics if the key types in both semanticIds are identical as well plus the other attributes of a Reference like _referredSemanticId_. So if the first key type of property "ManufacturerName" is "ExternalReference" and the first key type of the semanticId of property  "Herstellername" is "ModelReference" then the two semanticIds are no exact match. Or if the first reference contains a referredSemanticId value and the second does not the two semantic IDs are not exact matches.
+
```
BirgitBoss, line 456:
```suggestion
====
Typically, a semanticId does not have a _referredSemanticId_
====
```

documentation/IDTA-01001/modules/ROOT/pages/general.adoc
```diff
@@ -510,13 +551,15 @@ Matching does not mean to define equivalence classes that allow to overwrite con
 (GlobalReference)0173-1#01-ADS698#010, (GlobalReference)0173-1#01-ADS700#010
 ....
 
-matches
+is not exact match for
```
BirgitBoss, line 554:
```suggestion
is no exact match for
```

documentation/IDTA-01001/modules/ROOT/pages/general.adoc
BirgitBoss: We can add type and referredSemanticid

BirgitBoss:
@s-heppner Please check again whether it is ok now

s-heppner (approved):

PR #486, opened 2024-11-19
No recommendation for external references for semantic identifier

### LINKED ISSUE #376: HasSemantics/semanticId shall always be an External Reference

### LINKED ISSUE #396: Handle valueId and isCaseOf like semanticId
In https://github.com/admin-shell-io/aas-specs/issues/376 it was decided that all semanticIds should be external references.

valueId as defined for Property and MultiLanguageProperty are also referencing a concept definition. This is why also here text should be added that they are handled like semanticIds with the same constraints and recommendations etc.

isCaseOf is also a reference to a concept definition and therefore should also be handled like semanticId

BirgitBoss:
closes #376
closes #396
semantic identifier for semanticId, valueId, isCaseOf no recommendation any longer to use external references


BirgitBoss:
@J-Blume: may you please have a look? Thank you

s-heppner (approved):

PR #502, opened 2024-12-13
editorial changes constraints

BirgitBoss:
constraints are now displayed but they should not, removed \ for new text line


BirgitBoss:
wrong branch, changing to IDTA-01001-3-1_working does not work

PR #529, opened 2025-02-24
Add data type "DateTimeUtc"

### LINKED ISSUE #498: [BUG] Regex for xs:dateTime does not properly support timezones
**Describe the bug**
The regex for values of type xs:dateType 
```
^-?(([1-9][0-9][0-9][0-9]+)|(0[0-9][0-9][0-9]))-((0[1-9])|(1[0-2]))-((0[1-9])|([12][0-9])|(3[01]))T(((([01][0-9])|(2[0-3])):[0-5][0-9]:([0-5][0-9])(\\.[0-9]+)?)|24:00:00(\\.0+)?)(Z|\\+00:00|-00:00)
```
does only allow `Z`, `+00:00`, or `-00:00` as timezones.
Correct behavior would be to allow `Z` or any other timezone of format `(+|-)[0-9]{2}:[0-9]{2}`.

**Where**

- https://github.com/admin-shell-io/aas-specs/blob/04a1f04b9748978767c21d765cb464249a818d2f/schemas/json/aas.json#L224
- https://github.com/admin-shell-io/aas-specs/blob/04a1f04b9748978767c21d765cb464249a818d2f/schemas/json/aas.json#L622

**Additional context**
https://www.w3.org/TR/xmlschema-2/#dateTime Section 3.2.7.3

BirgitBoss:
closes #498
+ bugfix negative integer example


BirgitBoss:
@s-heppner should be conformant to what is already implemented in schema, correct?

s-heppner:
As far as I can see, it would not change the resulting schema files, however it will need adaption of aas-core-meta with the new type.

PR #536, opened 2025-03-07
Corrections

BirgitBoss:
move AasSubmodelElements to normative part, is used as type
correct figures
correct valid reference example (was correct before, now better explained)
remove link to non-existing Annex
improve referencing of constraints


JoergNeidig (approved):
I approve the proposed changes

PR #541, opened 2025-03-15
bugfix MessageTopicType is max 255 et.al.

BirgitBoss:


sebbader-sap (approved):
Will there be a note in the ChangeLog as well?

BirgitBoss:
> Will there be a note in the ChangeLog as well?

No, because it was correct in previous versions, I do not know why it was changed here in the first place

Issue #540, opened 2025-03-14
Inconsistent Length Restrictions for File/value and FileValue/value


sebbader-sap:
**Describe the bug**

File/value has an upper limit of 2048: https://github.com/admin-shell-io/aas-specs/blob/4a56e75e223681e6782b3a7da952bbd30301f67c/schemas/json/aas.json#L675

FileValue/value has an upper limit of 2000: https://github.com/admin-shell-io/aas-specs/blob/4a56e75e223681e6782b3a7da952bbd30301f67c/documentation/IDTA-01001/modules/ROOT/pages/mappings/encodings/valueonly-json-schema.adoc?plain=1#L117

**Solution**

Change maxLength for FileValue/value to 2048.

sebbader-sap:
I already use the 2048 in the OpenAPI declaration of FileValue/value :-)

BirgitBoss:
Thanks, also maxLength of contentType needs to be fixed: was changed from 100 to 128

BirgitBoss:
fixed with https://github.com/admin-shell-io/aas-specs/pull/543

PR #543, opened 2025-03-24
closes Bug #540 maxLength FileValue is 2048 and maxLength contentType is 128

BirgitBoss:
closes Bug #540 
* maxLength FileValue is 2048 and 
* maxLength contentType is 128

+ remove special char in file (editorial only)


JoergNeidig (approved):
The changes are straight-forward. Approve.

PR #573, opened 2025-05-23
Include the text of events.adoc in general.adoc

juileetikekar:


BirgitBoss (changes requested):
in this case please remove the corresponding event.adoc in includes

juileetikekar:
We don't need this PR anymore, as the issue has been fixed by #575

Issue #585, opened 2025-05-30
[Question] Why the sequence of elements inside embeddedDataSpecification changed from previous version 3.0?


mdanish98:
I am comparing the changes between XML schemas from 3.0 and 3.1.0 (latest released) and I am not sure why the sequence of elements inside embeddedDataSpecification changed from previous version 3.0. 

I went through the changes specified in the [Part 1](https://industrialdigitaltwin.org/en/wp-content/uploads/sites/2/2025/03/IDTA-01001-3-0-2_SpecificationAssetAdministrationShell_Part1_Metamodel.pdf) but couldn't find any relevant references.

Version 3.0:
```
<xs:group name="embeddedDataSpecification">
	<xs:sequence>
		<xs:element name="dataSpecification" type="reference_t"/>
		<xs:element name="dataSpecificationContent">
			<xs:complexType>
				<xs:sequence>
					<xs:group ref="dataSpecificationContent_choice"/>
				</xs:sequence>
			</xs:complexType>
		</xs:element>
	</xs:sequence>
</xs:group>
```

Version 3.1.0:
```
<xs:group name="embeddedDataSpecification">
    <xs:sequence>
      <xs:element name="dataSpecificationContent">
        <xs:complexType>
          <xs:sequence>
            <xs:group ref="dataSpecificationContent_choice"/>
          </xs:sequence>
        </xs:complexType>
      </xs:element>
      <xs:element name="dataSpecification" type="reference_t"/>
    </xs:sequence>
  </xs:group>
```

Could you please let me know any reference or reason why dataSpecificationContent and dataSpecification sequence is interchanged?

mristin:
I'm not 100% sure what caused this change -- but I suspect it has something to do with the order of optional/required elements (required coming first, before the optional ones).

@s-heppner do you perhaps know what happened here?

sebbader-sap:
>Lastly, @mdanish98 the schema files are generated and therefore cannot be edited manually, so thank you for your pull request, though the issue also persists in JSON and RDF, so I'll have to fix it in aas-core.

Just for completeness: JSON and RDF don't bother about the sequence, that's only a topic for XML. So also only XML is affected here.

empwilli:
> It's true, that it doesn't matter logically, but the documents will change when I change the order in aas-core, as all schema files are generated from the same source.

[...]

PR #586, opened 2025-06-02
Reverts the sequence of elements in embeddedDataSpecification of XML schema

mdanish98:
What changed:
Reverted the order of elements inside the embeddedDataSpecification group to match the correct sequence from version 3.0 — placing dataSpecification before dataSpecificationContent.

Why:
I think the element order was unintentionally swapped in the 3.1.0 schema, which deviated from the intended structure defined in version 3.0. This update corrects that inconsistency and restores the proper sequence as per the original specification logic.

Issue: #585 

I created this PR for the discussion and this PR can be further extended based on the outcomes of the discussion.

### DIFF
--- a/schemas/xml/AAS.xsd
+++ b/schemas/xml/AAS.xsd
@@ -277,6 +277,7 @@
   </xs:group>
   <xs:group name="embeddedDataSpecification">
     <xs:sequence>
+      <xs:element name="dataSpecification" type="reference_t"/>
       <xs:element name="dataSpecificationContent">
         <xs:complexType>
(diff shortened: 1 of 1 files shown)


s-heppner:
Thank you for your contribution! However, the schema files are generated and therefore cannot be edited manually (as then we would have to redo all manual changes every time we generate a new version). 

Due to this bug (see my comment in #585) we need to regenerate all JSON, XML, RDF and the suiting example files. 

I'll take care of this issue over the course of this week, so I will close your PR.

PR #646, opened 2025-12-24
Patch 2

praful-hunde:
All links in nav.adoc are missing pages folder in the path . All links are corrected.


BirgitBoss:
Hi @praful-hunde thank you for your  careful reviewing. In this case it is [antora ](https://antora.org/)that is using "pages" as its default, therefore it does not need to be added. I am not sure it would work if changing it.

Did you recognize any broken links in https://industrialdigitaltwin.io/aas-specifications/IDTA-01001/v3.2/index.html ?

I will close the PR without merging, if you have any converns please feel free to reopen it or add an issue. Thank you very much!

PR #669, opened 2026-03-11
remove infinite conditions for SubmodelElement Range

### LINKED ISSUE #639: Clarification on Range in ValueOnly representation
In [Format "Value" (Value-Only Serialization) in JSON
](https://industrialdigitaltwin.io/aas-specifications/IDTA-01001/v3.1.1/mappings/mappings.html#json-schema-value-only) it says:

> SubmodelElements without a value are not serialized.

Further below in the section about Ranges, it details more specifically:

> Range is serialized as named JSON object with ${Range/idShort} as the name of the containing JSON property. The JSON object contains two JSON properties. The first is named "min". The second is named "max". Their corresponding values are ${Range/min} and ${Range/max}.

To me, this reads, that Range objects in ValueOnly representation _always_ have 2 properties: "min" and "max". However, when checking the attributes of Range in the [submodel elements](https://industrialdigitaltwin.io/aas-specifications/IDTA-01001/v3.1.1/spec-metamodel/submodel-elements.html#range-attributes), we see that "min" and "max" are optional, with the following additional notes:

> If the min value is missing, the value is assumed to be negative infinite.
> If the max value is missing, the value is assumed to be positive infinite.

This raises the following questions:

- Are `{"min": "123"}` and `{"max": "123"}` valid Value Only representations of a Range Submodel Element? (As opposed to the paragraph above where it says, that "min" and "max" should be set?) (This appears to be the case due to the formulation: "Optional elements (like for example globalAssetId for an Entity submodel element) with no value shall be omitted in the serialization.")
- Is it valid to model a Range Submodel Element with unset "min" and "max", i.e., following the spec, a range from negative infinite to positive infinite?
- If the former is valid, what does the respective representation in Value Only representation look like?

Can you please clarify these questions?

BirgitBoss:
closes #639


documentation/IDTA-01001/modules/ROOT/pages/changelog.adoc
```diff
@@ -44,6 +44,8 @@ Major Changes:
 
 * CHANGED: new attributes for creation and update date in AdministrationInformation  (https://github.com/admin-shell-io/aas-specs-metamodel/issues/484[#484])
 * CHANGED: new value to enumeration AssetKind: Batch to support Digital Product Passports (https://github.com/admin-shell-io/aas-specs-metamodel/issues/483[#483])
+* CHANGED: SubmodelElement "Range" - remove conditions "If the min value is missing, the value is assumed to be negative infinite." and "If the max value is missing, the value is assumed to be positive infinite." 
+  (https://github.com/admin-shell-io/aas-specs-metamodel/issues/639[#639])
```
JoergNeidig, line 48: Suggestion to add "Range is still experimental" in the changelog entry.

JoergNeidig (approved):
Approved, with one suggestion

documentation/IDTA-01001/modules/ROOT/pages/changelog.adoc
BirgitBoss: > Suggestion to add "Range is still experimental" in the changelog entry.

It mentioned in the table in the changelog, so this should be sufficient, no?

PR #672, opened 2026-03-23
recursion depth and Value-Only Lists in Lists

### LINKED ISSUE #333: Limit recursion depth for SMEs
There should be a decision on a maximal recursion depth of SubmodelElements. Some SubmodelElements may contain arbitrary other SubmodelElements - namely:
- SubmodelElementList
- SubmodelElementCollection
- Entity
- AnnotatedRelationshipElement

This prevents overflow errors and should also serve as a guideline for modeling. 

**Describe the solution you'd like**
Define a maximal recursion depth in the spec. Suggestion: 100.

**Describe alternatives you've considered**
Leave everything as is. There probably won't be any problems until someone has a really complex scenario where their stack crashes at runtime.

### LINKED ISSUE #644: [QUESTION] What is a cannonical representation for SubmodelElementLists with property value = null?
According to meta model, submodel element lists may have a value = null, as value has a cardinality of 0..* (https://industrialdigitaltwin.io/aas-specifications/IDTA-01001/v3.1/spec-metamodel/submodel-elements.html#submodel-element-list-attributes) and the spec additionally forbids e.g. [empty json lists](https://github.com/admin-shell-io/aas-specs-metamodel/blob/0be69759372d8cd50f98850f9cc9ea8ce200455d/documentation/IDTA-01001/modules/ROOT/pages/mappings/formats/json.adoc). Same holds for XML.

This raises two questions:

1. For Submodel Instances: Consider a SMT with an optional SubmodelElementList, that may be empty, i.e., a SubmodelElementList with Qualifier SMT/Cardinality of value ZeroToOne with list elements with Qualifier SMT/Cardinality of value ZeroToMany.

If in my Submodel Instance this list is empty. Should I skip the SubmodelElementList element altogether from my submodel, or should I include it, but set its "value" to null?

 2. For Submodel Templates: given, that SubmodelElementLists include significant descriptive information about its item type (valueTypeListElement, typeValueListElement, semanticIdListElement), do I always have to specify a list item in my SMT or not? If I not, is there a "default" SMT/Cardinality qualifier supposed for the items?

BirgitBoss:
closes #333 
closes #644 

ADDED: A minimum recursion depth of 32 for Container-Elements should be supported. Note: This means for certification a maximum of 32 recursion test cases should be tested.

ADDED: Example and Note for Value-Only for SubmodelElementLists in SubmodelElementList: empty lists included


JoergNeidig (approved):
changed in accordance with workstream decision

PR #673, opened 2026-03-23
#462 SME Operation

### LINKED ISSUE #462: Modelling of operation parameters
Image the following scenario: We have an operation called `doStuff` which takes a single input parameter called `data`. `data` is a SubmodelElementList of SubmodelElementCollections with a predefined structure, i.e. Property `x` of type string, Property `y` of type integer, and Property `z` of type boolean.

An example invocation call with valueOnly serialization would look like this
````JSON
{
	"inputArguments":
	{
		
		"data": [
			{
				"x": "value1",
				"y": 0,
				"z": true
			},
			{
				"x": "value2",
				"y": 1,
				"z": false
			},
			...
		]
	},
	"inoutputArguments": {},
	"clientTimeoutDuration": "PT60S"
}
````

How should the operation be modeled to reflect this? If I model the `data` input parameter as a list with one element to describe the structure of the SubmodelElementCollection that is expected like this...

```JSON

{
	"modelType": "Operation",
	"idShort": "doStuff"
	"inputVariables": [
		{
			"value":
			{
				"modelType": "SubmodelElementList",
				"idShort": "data",
				"orderRelevant": true,
				"typeValueListElement": "SubmodelElementCollection",
				"value": [
					{
						"modelType": "SubmodelElementCollection",
						"value": [
							{
								"modelType": "Property",
								"idShort": "x",
								"valueType": "xs:string"
							},
							{
								"modelType": "Property",
								"idShort": "y",
								"valueType": "xs:integer"
							},
							{
								"modelType": "Property",
								"idShort": "z",
								"valueType": "xs:boolean"
							},
						]
					}
				]
			}
		}
	],
	"outputVariables": []
}
```

... then this reads like `data` only accepts exactly one element instead of an arbitrary number. If you argue that this should mean that `data` contains an arbitrary number of elements than how is it possible to define that it should contain exactly one element?

I was looking at Part 1 of the AAS specification to find an answer but unfortunately without success. 

My initial thought was that setting the `kind` property on the SubmodelElementCollection to `template` might be the answer, but this property has been marked as deprecated sin v3.0.

Using Qualifiers also came to mind but it seems TemplateQualifiers are only allowed on elements that belong to a Submodel of `kind=template` (which is not the case here is this Submodel is an instance) according to `Constraint AASd-129`

>If any Qualifier/kind value of a Qualifiable/qualifier is equal to TemplateQualifier and the qualified element inherits from "hasKind", the qualified element shall be of kind Template (HasKind/kind = "Template").

To me, it seems like I am therefore not able to properly model/describe the input parameters of my operation which means nobody can use my AAS/operation without an accompanying document describing how to use it.
Am I missing something? Is there a way to describe this?

BirgitBoss:
closes  #462

* remove Note 3 of Operation
* add that OperationVariable handled like Template also in Submodel Instance
* new Constraint AASd-138 
* rename Constraint AASd-135 to AASd-137 because already used before (although not contained any longer)
* editorial changes


documentation/IDTA-01001/modules/ROOT/pages/includes/constraints.adoc
```diff
@@ -41,9 +41,10 @@ every entry needs to be in one line, no \
...
-:aasd135: pass:q[[underline]#Constraint AASd-135:# For external references, i.e. xref:ROOT:spec-metamodel/referencing.adoc#Reference[Reference] the value of any xref:ROOT:spec-metamodel/referencing.adoc#Key[Key/type] contained in the Reference shall not be one of xref:ROOT:spec-metamodel/referencing.adoc#AasReferables[AasReferables].]
\ No newline at end of file
+:aasd137: pass:q[[underline]#Constraint AASd-137:# For external references, i.e. xref:ROOT:spec-metamodel/referencing.adoc#Reference[Reference] the value of any xref:ROOT:spec-metamodel/referencing.adoc#Key[Key/type] contained in the Reference shall not be one of xref:ROOT:spec-metamodel/referencing.adoc#AasReferables[AasReferables].]
+:aasd138: pass:q[[underline]#Constraint AASd-138:# A xref:ROOT:spec-metamodel/submodel-elements.adoc#SubmodelElementList[SubmodelElementList] within a xref:ROOT:spec-metamodel/core.adoc#Submodel[Submodel] of xref:spec-metamodel/common.adoc#HasKind[kind]=xref:spec-metamodel/common.adoc#ModellingKind[Template] or as part of an xref:ROOT:spec-metamodel/submodel-elements.adoc#Operation[OperationVariable] shall have exactly one element.]
```
mjacoby, line 50: Would it make sense to enforce that the list must have a qualifier to indicate the cardinality of the elements inside the list? At least I cannot think of any example where cardinality is not relevant. To reduce overhead and to keep backwards capability, we could additionally define a default cardinality in case no cardinality is explicitly given, e.g. `0 .. *`

documentation/IDTA-01001/modules/ROOT/pages/spec-metamodel/submodel-elements.adoc
```diff
@@ -621,19 +621,17 @@ h|Explanation h|Type h|Card.
 a|Describes an argument or result of an operation via a SubmodelElement |xref:spec-metamodel/core.adoc#SubmodelElement[SubmodelElement] |1
 |===
 
-====
-Note 1: an operation can  be invoked via an API call. For further explanation see Part 2 (IDTA-01002).
-====
+Handling of operations in Submodel with xref:spec-metamodel/common.adoc#HasKind[kind]=xref:spec-metamodel/common.adoc#ModellingKind[Instance] are handled identical to operation in Submodel Templates (Submodel with xref:spec-metamodel/common.adoc#HasKind[kind]=xref:spec-metamodel/common.adoc#ModellingKind[Template])
```
mjacoby:
```suggestion
Handling of operations in Submodel with xref:spec-metamodel/common.adoc#HasKind[kind]=xref:spec-metamodel/common.adoc#ModellingKind[Instance] are handled identical to operations in Submodel Templates (Submodel with xref:spec-metamodel/common.adoc#HasKind[kind]=xref:spec-metamodel/common.adoc#ModellingKind[Template])
```

Also, not sure if operations on submodels with `kind=template` should be invokable but also no strong opinion. For me, this is fix and only the typo needs to be fixed.

documentation/IDTA-01001/modules/ROOT/pages/changelog.adoc
```diff
@@ -127,12 +131,23 @@ use "SubmodelElement" instead of "submodel element"
 use "SubmodelElement" instead of "submodel element" 
 
 |{empty}  |AASd-129 |Update a|
-use "SubmodelElement" instead of "submodel element" and "Submodel" instead of "submodel"
+* add "Exception: the SubmodelElement is part of an OperationVariable." 
+* use "SubmodelElement" instead of "submodel element" and "Submodel" instead of "submodel"
 
-|(x)  |AASd-135 |New a|
+Constraint AASd-129: If any Qualifier/kind value of a SubmodelElement/qualifier (attribute qualifier inherited via Qualifiable) is equal to TemplateQualifier, 
+the SubmodelElement shall be part of a Submodel template,
```
mjacoby: Unintended line break at end of this line?

mjacoby (commented):
In general I approve the proposed changes but there are some minor spelling/formatting issues and some proposals to maybe improve the content even further.

documentation/IDTA-01001/modules/ROOT/pages/includes/constraints.adoc
BirgitBoss: So far we did not introduce the cardinality qualifier etc. in the AAS Specifications, only in the How-to create a SMT. Therefore I think we cannot (yet) add it.

documentation/IDTA-01001/modules/ROOT/pages/changelog.adoc
BirgitBoss:
```suggestion
Constraint AASd-129: If any Qualifier/kind value of a SubmodelElement/qualifier (attribute qualifier inherited via Qualifiable) is equal to TemplateQualifier, the SubmodelElement shall be part of a Submodel template, 
```

documentation/IDTA-01001/modules/ROOT/pages/spec-metamodel/submodel-elements.adoc
```diff
@@ -621,19 +621,17 @@ h|Explanation h|Type h|Card.
 a|Describes an argument or result of an operation via a SubmodelElement |xref:spec-metamodel/core.adoc#SubmodelElement[SubmodelElement] |1
 |===
 
-====
-Note 1: an operation can  be invoked via an API call. For further explanation see Part 2 (IDTA-01002).
-====
+Handling of operations in Submodel with xref:spec-metamodel/common.adoc#HasKind[kind]=xref:spec-metamodel/common.adoc#ModellingKind[Instance] are handled identical to operations in Submodel Templates (Submodel with xref:spec-metamodel/common.adoc#HasKind[kind]=xref:spec-metamodel/common.adoc#ModellingKind[Template])
```
BirgitBoss:
```suggestion
Specification of operations in Submodel with xref:spec-metamodel/common.adoc#HasKind[kind]=xref:spec-metamodel/common.adoc#ModellingKind[Instance] are handled identical to operations in Submodel Templates (Submodel with xref:spec-metamodel/common.adoc#HasKind[kind]=xref:spec-metamodel/common.adoc#ModellingKind[Template])
```

documentation/IDTA-01001/modules/ROOT/pages/spec-metamodel/submodel-elements.adoc
BirgitBoss: I changed it to "specification" instead of "handling", operations in templates from my point of view are not invokable

PR #689, opened 2026-04-17
docs(terms): add cross-spec terminology matrix (IDTA-01002/01004)

aorzelskiGH:
## Summary

Add a normative cross-specification terminology matrix at the end of the "Terms, Definitions and Abbreviations" section. The table maps every top-level AAS entity to its API resource / path segment (IDTA-01002) and its Query / Access-Rule FieldIdentifier prefix (IDTA-01002 Query Language and IDTA-01004 Access Rule Model).

## Problem

The three specifications use different surface forms for the same concept:

- Metamodel uses class names (`AssetAdministrationShell`, `Submodel`, ...).
- API uses URL path segments (`/shells`, `/submodels`, ...).
- Query / Access Rules use lower-case prefixes (`$aas`, `$sm`, `$sme`, `$cd`, `$aasdesc`, `$smdesc`).

Readers who only work with one spec keep producing small inconsistencies (e.g. `$aasDesc` vs. `$aasdesc`, "Shell" vs. "AAS"), and new editors have to rediscover the mapping each review cycle.

## Solution

Add a single cross-reference table to the Metamodel's Terms section. The API and Security specs can reference this table instead of duplicating the mapping. The table also states that prefixes are lower-case and that variants like `$aasDesc` MUST be rejected by parsers.

## Affected files

- `documentation/IDTA-01001/modules/ROOT/pages/terms-definitions-and-abbreviations.adoc`

## Review notes

- No normative changes to the Metamodel itself; this is a cross-spec clarification.
- API PR (to follow) and Security PR (to follow) will reference this section.

Refs: Review Finding T-09

### DIFF
--- a/documentation/IDTA-01001/modules/ROOT/pages/terms-definitions-and-abbreviations.adoc
+++ b/documentation/IDTA-01001/modules/ROOT/pages/terms-definitions-and-abbreviations.adoc
@@ -471,3 +471,45 @@ The following abbreviations are not used in the document but may be used as abbr
 |SME |SubmodelElement
 |SML |SubmodelElementList
 |===
+
+[#cross-spec-terminology]
+== Terminology Across IDTA-01001, IDTA-01002 and IDTA-01004
+
+The following table relates the core AAS concepts across the three specifications. It is intended as a quick cross-reference for readers of the API (IDTA-01002) and Security (IDTA-01004) specifications. All three specifications use the same underlying entities; only their textual representations differ (metamodel class name, HTTP path segment, query / access-rule FieldIdentifier prefix).
+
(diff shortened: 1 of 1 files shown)


BirgitBoss (commented):
so far in Part 1 the terms "AssetAdministrationShellDescriptor" and "SubmodelDescriptor" were not introduced

BirgitBoss:
I propose to not accept this PR. The content can be added to API as Annex.

BirgitBoss:
TF Metamodel 2026-06-19: do not merge into V3.2. Unnecessary dependency to API and Security. However, table is seen as very helpful and should be added to at least Part 2.

PR #690, opened 2026-04-17
docs(index): add cross-spec alignment matrix (T-08)

aorzelskiGH:
## Summary

Add an explicit "Alignment with other AAS Specifications" table to the Metamodel's index page. The table names the companion IDTA-01002 (API, v3.2) and IDTA-01004 (Security, v3.1) versions this Metamodel version is designed to be used with, and clarifies the substitution rule for bugfix (patch) releases.

## Problem

IDTA-01001 currently only lists dependencies to IDTA-01003-a/b (data specifications). It does not state which versions of the API (IDTA-01002) and Security (IDTA-01004) specifications this Metamodel version is aligned with. Reviewers and implementers have to infer version compatibility from the other two specs, which is error-prone.

## Solution

- Add a short normative `=== Alignment with other AAS Specifications` section directly below `=== Dependencies` in `index.adoc`.
- Present the information as a simple table (Specification / Aligned Version / Notes).
- Keep the existing "If there is a bugfix of these parts, they shall be used" rule and extend it to apply to aligned specs as well.

## Affected files

- `documentation/IDTA-01001/modules/ROOT/pages/index.adoc`

## Review notes

- No changes to the metamodel itself.
- Companion PRs in `aas-specs-api` and `aas-specs-security` add mirroring alignment tables.

Refs: Review Finding T-08

### DIFF
--- a/documentation/IDTA-01001/modules/ROOT/pages/index.adoc
+++ b/documentation/IDTA-01001/modules/ROOT/pages/index.adoc
@@ -26,11 +26,11 @@ Previous version: 3.1.2
 [#cross-spec-alignment]
 === Alignment with other AAS Specifications
 
-This version of IDTA-01002 (v3.2) is designed to be used together with the following companion specifications:
+This version of IDTA-01002 (v3.2) is designed to be used together with the following aligned companion specifications:
 
 [cols="1,2,1,3",options="header"]
 |===
-h| Dependency h| Specification h| Aligned Version h| Notes
+h| Dependency h| Specification h| Aligned Version h| Notes, IDTA-01001 ...
 
 |  | IDTA-01002 Part 2: Application Programming Interfaces
 | v3.2
@@ -69,6 +69,8 @@ a|
 |===
 
 Bugfix (patch) releases of aligned specifications shall be substituted. 
+Minor versions of aligned specifications IDTA-01001 does not have a strong dependency to
+ may be used as long as the corresponding release version of the aligned specification does not have a strong dependency to a newer release of IDTA-01001.
 
 
 == Notice
--- a/documentation/IDTA-01001/modules/ROOT/pages/index.adoc
+++ b/documentation/IDTA-01001/modules/ROOT/pages/index.adoc
@@ -31,6 +31,26 @@ For the schemas derived from this document there is a dependency to the followin
 
 If there is a bugfix of these parts, they shall be used.
 
+[#cross-spec-alignment]
+=== Alignment with other AAS Specifications
+
+This version of IDTA-01001 (Metamodel, v3.2) is designed to be used together with the following companion specifications:
+
+[cols="2,2,3",options="header"]
+|===
+| Specification | Aligned Version | Notes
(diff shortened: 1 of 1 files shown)


BirgitBoss:
In the past we explicitly did not do an alignment with the release in case a part does not have an update. So for example, if only the API is extended and gets a new version but the metamodel remains the same with this newly proposed section an unnecessary dependency is introduced.

documentation/IDTA-01001/modules/ROOT/pages/index.adoc
```diff
@@ -31,6 +31,26 @@ For the schemas derived from this document there is a dependency to the followin
 
 If there is a bugfix of these parts, they shall be used.
```
BirgitBoss: This chapter has the same intention as the newly introduced chapter "Alignment with other AAS specifications". The new proposal add more  information on what to expect in which version. On of both must be removed.

documentation/IDTA-01001/modules/ROOT/pages/index.adoc
```diff
@@ -31,6 +31,26 @@ For the schemas derived from this document there is a dependency to the followin
...
+| v3.2
+| HTTP/REST API, AAS Query Language. References the Metamodel classes and attributes defined here.
+
+| IDTA-01004 Part 4: Security
+| v3.1
+| Access Rule Model. Shares the Query Language grammar and schema with IDTA-01002 and references Metamodel classes through FieldIdentifiers.
+|===
+
+Any bugfix release (patch version) of an aligned specification MAY be substituted for the version listed above. Implementations claiming conformance to this version of IDTA-01001 SHOULD state which versions of IDTA-01002 and IDTA-01004 they implement.
+
```
BirgitBoss, line 74: This chapter has the same intention as the existing chapter "Dependencies". The new proposal adds more  information on what to expect in which version. One of both must be removed.

BirgitBoss (changes requested):

documentation/IDTA-01001/modules/ROOT/pages/index.adoc
```diff
@@ -21,15 +21,57 @@ This is version 3.2 of the specification IDTA-01001.
...
+|  | IDTA-01005 Part 5: Package File Format (AASX)
+| v3.2
+a| 
+
+* defines mapping rules for serialization format(s) of the exchanged Asset Administration Shell referenced via `aasx-origin` 
+|===
+
+Bugfix (patch) releases of aligned specifications shall be substituted. 
+Minor versions of aligned specifications IDTA-01001 does not have a strong dependency to
+ may be used as long as the corresponding release version of the aligned specification does not have a strong dependency to a newer release of IDTA-01001.
```
BirgitBoss:
```suggestion
Minor versions of aligned specifications IDTA-01001 may be used as long as the corresponding release version of the aligned specification does not have a dependency to a newer release of IDTA-01001.
```

BirgitBoss (approved):

PR #715, opened 2026-08-27
Add patch version to administrativeInformation

### LINKED ISSUE #684: Add Patch version to AdministrativeInformation
We have version and revision but no patch version. Since version and revision are numbers only, it is not possible to describe a version 1.0.1, for example for bugfixes.

**Describe the solution you'd like**

Add optional attribute 
  ` patch`
to AdministrativeInformation with string data type and rule stating that patches when sorted shall be sortable according time of creation.

I.e.
If <version>.<revision>.<patch>  < <version>.<revision>.<patch> 
then
createdAt and updatedAt of first Identifiable <  updatedAt of second Identibiable (both having the same ID) (the createdAt of second shall be identical to the first one, otherwise they should have different IDs.

JoergNeidig:
- Add patch attribute to AdministrativeInformation
- Add patchType to DataTypes
- Changed administrative-information.puml to include patch
- Change AASd-005
- Update Changelog

closes #684


documentation/IDTA-01001/modules/ROOT/pages/spec-metamodel/datatypes.adoc
```diff
@@ -234,7 +234,16 @@ Each string within langString has a max of 1,023 and min of 1 characters.
 .2+e|[[NameType]]NameType 2+| `\https://admin-shell.io/aas/3/2/NameType`
 a|_string_ with max 128 and min 1 characters |"ManufacturerPartId"
 
+.2+e|[[PatchType]]PatchType 2+| `\https://admin-shell.io/aas/3/2/PatchType`
```
BirgitBoss, line 237: I think the Patch Type differs from the version and revision type, it should start with a number but then also Text can be added. see https://semver.org/spec/v2.0.0.html. The patch in SEMVER is just a number, but otherwise we would need to add additional fielda for the release- or build-suffix etc.?

^(0|[1-9][0-9]*)([-+].*)?$

examples:
2
1-alpha.1
2+20130313144700

documentation/IDTA-01001/modules/ROOT/pages/changelog.adoc
```diff
@@ -70,6 +74,9 @@ h|Nc h|V3.3 Change w.r.t. V3.2 h|Comment
 h|Nc h|New Elements V3.3 vs V3.2 h|Comment
 |{empty} | xref:spec-metamodel/submodel-elements.adoc#File[File/fingerprint]  a| new attribute with cardinality 0..*
 |{empty} | xref:spec-metamodel/datatypes.adoc#FingerprintType[FingerprintType]  a| new datatype for xref:spec-metamodel/submodel-elements.adoc#File[File/fingerprint]
+|{empty} | xref:spec-metamodel/common.adoc#AdministrativeInformation[AdministrativeInformation/patch]  a| new attribute with cardinality 0..*
```
BirgitBoss, line 77: cardinality is 0..1 but this information belongs to AdministrativeInformation not to patch attribute itself I would say

BirgitBoss (changes requested):

BirgitBoss:
We should also extend our example for sorting

```
This numerical system creates a clear and logical order of precedence, where a higher version or revision number always indicates a later release:

V1.0 < V1.1 < V2.0
```
to
`V1.0 < V1.1 < 1.1.2 < V2.0`

with text it is difficult,.... how to sort? or just ignore?

`V1.0 < V1.1-1gh < V1.1-abc <  < 1.1.2 < V2.0`

documentation/IDTA-01001/modules/ROOT/pages/spec-metamodel/datatypes.adoc
JoergNeidig: The current PR adds only patch. Patch is always numerical. 

We could also add "pre-release" (1.0.0-alpha, 1.0.0-alpha.1, 1.0.0-0.3.7, ...) and "Build" (1.0.0+20130313144700, 1.0.0+exp.sha.5114f85) which are both alphanumeric.

The advantage of adding both is that we are than 100% compliant with Semver, but I am not sure that pre-release and build would be used in AdministrativeInformation for AAS-Elements.

JoergNeidig:
> We should also extend our example for sorting

If we add pre-release in addition to patch the sorting is more complicated (build must be ignored), but Semver defines that in detail. 
Example: 1.0.0-alpha < 1.0.0-alpha.1 < 1.0.0-alpha.beta < 1.0.0-beta < 1.0.0-beta.2 < 1.0.0-beta.11 < 1.0.0-rc.1 < 1.0.0.

I would recommend to simply reference Semver:

